PR #5, opened 2023-08-01
add load sound tools

mklimko-perpetio:
LoadFromUrl requires permission like microphone (at least on mac) so maybe we should add this to readme file
<key>com.apple.security.network.client</key>
    <true/>

### DIFF
--- a/lib/src/tools/soloud_loading_tool.dart
+++ b/lib/src/tools/soloud_loading_tool.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter_soloud/flutter_soloud.dart';
 import 'package:flutter_soloud/src/utils/assets_manager.dart';
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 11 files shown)


lib/src/tools/soloud_loading_tool.dart
alnitak, line 1: Instead of throwing exceptions, I'd rather prefer to return a null value and do a check after the call, like in the other Soloud methods.  For example if an url doesn't exist, the app doesn't crash if the try catch is forgotten.
I am fixing this if you mind.

alnitak (approved):
Instead of throwing exceptions, I'd rather prefer to return a null value and do a check after the call, like in the other Soloud methods.  For example if an url doesn't exist, the app doesn't crash if the try catch is forgotten.
I am fixing this if you mind.

alnitak:
Instead of throwing exceptions, I'd rather prefer to return a null value and do a check after the call, like in the other Soloud methods.  For example if an url doesn't exist, the app doesn't crash if the try catch is forgotten.
I am fixing this if you mind.

Issue #26, opened 2024-02-25
fix: A lot of RAM is needed


synchronisator:
**Description**

I load a bunch of audiofiles.
Round about 35 mp3 files with round about 5MB per file a complete size of 180mb.
On Linux i can see, that the RAM usage goes ob from ~450MB to 3,1GB
On Android this leads to a shutdown of the app by the system

**Steps To Reproduce**

Load a lot of big files.

**Expected Behavior**

The RAM usage should not go up this high.

alnitak:
Hi @synchronisator, you are right. The sounds are loaded into memory to make them accessible as fast as possible, useful for game sounds.
When you load a file, the sound handle is persistent and that sound can be played without lags. When you dispose that handle, also the memory is freed. So I suggest you to dispose the `SoundProp` when it ends.

Actually [SoLoud::Wav.load()](https://solhsa.com/soloud/wav.html) function is used for which the doc states:
`For lengthy samples like background music, you may want to use SoLoud::WavStream instead. The Wav is all about speed, and always decodes the whole sample into memory at load time.`

I put this on my TODO list

alnitak:
About the huge memory allocated: the audio data in an mp3 file is compressed, while into the allocated memory goes the uncompressed audio data. Since a second of audio takes about 350KB of data, every minute will take about 20 MB of uncompressed data.

synchronisator:
Thanks for your answer. I make a practicing app for choir singers with multiple voices. So its essential that the files are played in perfect sync. So i think the Stream is not an option, but i will try.
And i want it to behave like a "normal" mediaplayer. 
By the way: SoLoud works great with audio_service so is can use all features like Mediaplayer-notifications with SoLoud. Thats very nice.
So i first will try to only have one song loaded. That will cause every song to load for a few seconds before starting, but thats maybe worth it.

alnitak:
Hi @synchronisator,

I have made some modifications that point to resolve this specific issue.
You can now choose to load sound files not into memory. Please, [read the PR](https://github.com/alnitak/flutter_soloud/pull/30) to see also an unavoidable *seek problem*.

I will be very grateful if you will find some time to try it!
You just need to momentarily comment this line in your `pubspec.yaml`:
`flutter_soloud: ^1.2.4`

and add this:
```
  flutter_soloud:
    git:
      url: https://github.com/alnitak/flutter_soloud.git
      ref: wavStream
```

Thank you!

synchronisator:
[...]

PR #30, opened 2024-02-29
 fix: A lot of RAM is needed #26 

alnitak:
#### Added `loadIntoMem` parameter to `Soloud.loadFile()`.

By default is set to `true` which means no breaking changes.
if `true` *[Soloud::wav](https://solhsa.com/soloud/wav.html)* will be used which loads all raw audio data into memory. This will be useful when loading short audio, ie for game sounds, mainly used to prevent gaps or lags when starting/seeking a sound (less CPU, more memory allocated).
Every second of a 44100Hz stereo sound takes about 350kB of memory, means that a minute is about 20 mB. Shouldn't be a problem nowadays, but it could be if a lot of audio files are loaded as the issue #26 states.

If `false`, [*Soloud::wavStream*](https://solhsa.com/soloud/wavstream.html) will be used and the audio data is loaded 
from the given file when it's needed (more CPU less memory allocated). But, using `false`, a drawback is evident when **seeking backward** using ie a slider with *mp3* files:  the audio position must be reset to the start and then the new seek position can be calculated. This produce a little lag when using a slider, so, on most situation, I think it's better to load audio into memory.

## Type of Change

- [x ] ✨ New feature (non-breaking change which adds functionality)
- [ x] 📝 Documentation


alnitak:
**Using `loadIntoMem=false` with MP3s lags when seeking.**

The MP3 frames do not have a fixed size, therefore the correct seeking position must be computed by reading all data in between the current position and the new seeking position. The lag to calculate it is more evident when seeking backward where it is necessary to rewind to the start position and calculate the new position.
This doesn't happen with FLACs, OGGs, and WAVs. So I disabled seeking functionality using this mode with only MP3s.

Loading an audio file using this mode **could be useful**, especially for background music (not ie with an audio player where a seeking slider is a must).

Anyways, the need to use `loadIntoMem=false` comes from issue #26 where an uncommon use-case of loading more the 30 MP3 audio files.

I don't know if this could cause problems for developers who will use this package even if well documented.

How do you feel about this @filiph ?

lib/src/flutter_soloud_bindings_ffi.dart
```diff
@@ -107,17 +107,28 @@ class FlutterSoLoudFfi {
...
+  /// the audio is short, ie for game sounds, mainly used to prevent
+  /// gaps or lags when starting a sound (less CPU, more memory allocated).
+  /// If false, the audio data is loaded from the given file when
+  /// needed (more CPU less memory allocated).
+  /// See the [seek] note problem when using [loadIntoMem] = false
+  /// [soundHash] return hash of the sound.
+  /// Returns [PlayerErrors.noError] if success.
+  ({PlayerErrors error, int soundHash}) loadFile(
+    String completeFileName,
+    bool loadIntoMem,
```
filiph, line 124: This is a small nit, but I'd prefer `loadIntoMemory`, at least in the developer-facing API.

lib/src/soloud.dart
```diff
@@ -148,10 +148,6 @@ class SoLoud {
       } else {
         debugIsolates('******** MAIN EVENT data: $data');
         if (data is StreamSoundEvent) {
-          debugPrint('@@@@@@@@@@@ SOUND EVENT: ${data.event}  '
-              'handle: ${data.handle}  '
-              'sound: ${data.sound}');
-
```
filiph, line 154: These are helpful. I'd just do this using `log` instead of `debugPrint`. If you leave this in, I'll convert it when I do my pass of the codebase (hopefully today, tomorrow or on Monday).

lib/src/soloud.dart
```diff
@@ -328,24 +324,33 @@ class SoLoud {
...
+  /// See the [seek] note problem when using [loadIntoMem] = false
+  /// Default is true.
   /// [completeFileName] the complete file path
   /// Returns PlayerErrors.noError if success and a new sound
   ///
   Future<({PlayerErrors error, SoundProps? sound})> loadFile(
-    String completeFileName,
-  ) async {
+    String completeFileName, {
+    bool loadIntoMem = true,
```
filiph, line 340: The nit about naming this `loadIntoMemory` applies here more than anywhere else — this is the developer-facing API, if I'm not mistaken.

lib/src/soloud.dart
```diff
@@ -751,11 +756,21 @@ class SoLoud {
...
-  /// Seek playing in seconds
-  ///
-  /// [time] the time to seek
+  /// Seek playing in [time] seconds
+  /// [time]
   /// [handle] the sound handle
   /// Returns [PlayerErrors.noError] if success
+  /// 
+  /// NOTE: when seeking an mp3 file loaded using `loadIntoMem`=false
+  /// the seek operation is not performed due to lags. This occurs because the
```
filiph, line 765: Would it make more sense to perform the seek, even with the (large) lag? Keep the warning in the API. We can also log a warning into console every time the developer does this. But if the only issue is really "just" the lag, it seems prohibitive to completely deny this to the developer.

For example, I might load a background music mp3, play it, then fade it out to make room for some other jingle, and then seek to a position in the background music mp3 before fading it in again.

filiph: On the other hand, if it's more code, or more brittle, to allow mp3 seeking in disk mode, then by all means let's just disallow it.

lib/src/soloud.dart
filiph: Actually, one other thing that comes to mind is to make this an enum?

```dart
enum LoadMode {
  /// Load and decompress the audio file into RAM.
  memory,

  /// Keep the file on disk and only load chunks as needed.
  disk,
}
```

Then we'd have:

```dart
loadFile('path/to/music.mp3', mode: LoadMode.disk);
```

The enum also gives us an easier path later, if we decide to add more options. For example, if we add something like `LoadMode.balanced` or whatever, the developer-facing API stays the same.

To be clear: on the inside, this should still be implemented with a `bool`. At least for now. It's just about what the developer uses.

example/lib/page_3d_audio.dart
```diff
@@ -203,9 +203,9 @@ class _Audio3DWidgetState extends State<Audio3DWidget>
 
   void updatePos(Offset delta, Duration timeStamp) {
     velX =
-        100 * delta.dx / (timeStamp.inMilliseconds - precTime.inMilliseconds);
+        -100 * delta.dx / (timeStamp.inMilliseconds - precTime.inMilliseconds);
```
filiph, line 206: Good to see this fixed!

lib/src/bindings_capture_ffi.dart
```diff
@@ -47,8 +47,8 @@ class FlutterCaptureFfi {
 
     int ndev = n_devices.value;
     for (int i = 0; i < ndev; i++) {
-      var s = devices.elementAt(i).value.ref.name.cast<Utf8>().toDartString();
-      var n = devices.elementAt(i).value.ref.isDefault;
+      var s = (devices + i).value.ref.name.cast<Utf8>().toDartString();
```
filiph, line 50: Good to see this addressed as well. The lint warning bothered me. :)

filiph (approved):
Yes, I think this is an excellent call — the use case of having a very long (5+ minutes) audio file (a music track or an ambience track) that doesn't need to be low-latency is very real. And keeping it in memory as a WAV is wasteful, especially on mobile. I'm glad this is being addressed.

I had some suggestions but otherwise, LGTM!

lib/src/flutter_soloud_bindings_ffi.dart
filiph: (See below for more options.)

lib/src/soloud.dart
alnitak: That was my dilemma. I agree to enable the seek and write warnings in the doc/comments and into console logs.

PR #33, opened 2024-03-08
Rename start/stopIsolate and init/stopEngine to initialize() and dispose()

filiph:
## Description

Makes use of standard Dart style of initializing (with a single method) and disposing (with `dispose()`).

The previous methods (such as `startIsolate`) are still in the codebase — they're just marked `Deprecated` and forward calls to the new names. This way, user code will still work, it will just generate analysis warnings.

I've also implemented an automatic fix via [Data-driven fixes](https://github.com/flutter/flutter/wiki/Data-driven-Fixes). This way, developers can apply the changes to their code with `dart fix --apply` or via Quick Fix interfaces in their IDEs:

Screenshot 2024-03-08 at 13 44 56

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


filiph:
Marked as DRAFT because, while working on this PR, I realized something.

There's the `initialize()` and `dispose()` which work pretty well and seem very idiomatic, I think.

But there are also the `startCapture()` and `stopCapture()` methods. Are these completely independent from the main ("player") initialization? Should we, for example, have `initializePlayer()` and `initializeCapture()`? Or do the capture methods need the player to be initialized?

I assume we don't want to automatically initialize the capture subsystem in `initialize()` (formerly known as `startIsolate()`), because many developers will _not_ be using the mic, and want to avoid asking for permissions, etc. But for simplicity, how about having something like `SoLoud.initialize(withCapture: true)`? The idea is that if you know you might want to capture audio, you probably know it in advance.

Or, we can keep it as is, and just let `startCapture` and `stopCapture` be. It throws me off, though, that for example in `page_hello_flutter.dart`, we have:

```dart
    SoLoud().dispose();
    SoLoud().stopCapture();  // <-- capture stopped _after_ dispose?
```

I guess my question to you, @alnitak, is: how "independent" is capture from the "main" soloud? Is it something we can start and stop independently on the rest of soloud? Should we have `SoLoudPlayer()` and `SoLoudCapture()`? What do you think?

alnitak:
Yes, the player and the capture are completely independent so having `initializePlayer()` and `initializeCapture()` should be coherent!
- the player uses Soloud lib just for outputting audio.
- the capture uses miniaudio to just capture from microphone
They are 2 separate things and I think the developers should make their choice to start one of them or both. So no need to initialize the player to use the capture.

I used
```dart
    SoLoud().dispose();
    SoLoud().stopCapture();  // <-- capture stopped _after_ dispose?
```
just because passing from a tab to another I have to be sure that all (player and capture) are stopped to eventually reinitialize in another tab.

> I guess my question to you, @alnitak, is: how "independent" is capture from the "main" soloud? Is it something we can start and stop independently on the rest of soloud? Should we have SoLoudPlayer() and SoLoudCapture()? What do you think?
This could be a very good idea to clarify their scopes! Also because in the near future we could think about a `SoLoudStream` or something like that, to send audio data to be played from Flutter.

alnitak (approved):

PR #38, opened 2024-03-09
Extract capture methods into their own class: SoLoudCapture

filiph:
## Description

Extract capture methods into their own class: `SoLoudCapture`.

Also, add `SoLoud.instance`, and mark `SoLoud()` as deprecated.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/soloud_capture.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// ```dart
+  /// // BAD
+  /// var sound = await SoLoudCapture().loadFile('path/to/sound.mp3');
+  /// await SoLoudCapture().play(sound)
+  /// ```
+  ///
+  /// The code above suggests, on the face of it, that we're _constructing_
+  /// two instances of [SoLoudCapture], although we're in fact only accessing
+  /// the singleton instance.
+  @Deprecated('Use SoLoudCapture.instance instead')
```
alnitak, line 28: Since we have not yet published the plugin with `SoLoudCapture`, are these and the other deprecated advice necessary?

alnitak (commented):
Since we have not yet published the plugin with `SoLoudCapture`, are those `@deprecated` advice necessary?

lib/src/soloud_capture.dart
filiph: This one isn't _but_ it help when migrating. For example, you have code like this:

```dart
SoLoud().initCapture();
```

You get a lint warning to change to `SoLoudCapture()`. You do that. Now you get another warning to change to `SoLoudCapture.instance`. 

Compare this to when you have `SoLoud().initCapture()`, you try to change it to `SoLoudCapture()`, but that doesn't exist at all.

PR #42, opened 2024-03-11
Fix issue where errors from SoLoud would not propagate

filiph:
## Description

The C++ `play` and `play3d` methods return a 0 when there's an error. I neglected to convert that 
to `SoundHandle.error()`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/bindings_player_ffi.dart
+++ b/lib/src/bindings_player_ffi.dart
@@ -375,7 +375,8 @@ class FlutterSoLoudFfi {
   /// [volume] 1.0f full volume
   /// [pan] 0.0f centered
   /// [paused] 0 not pause
-  /// Return the handle of the sound, 0 if error
+  ///
+  /// Return the handle of the sound, [SoundHandle.error] if error
   SoundHandle play(
     SoundHash soundHash, {
     double volume = 1,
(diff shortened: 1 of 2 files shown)


lib/src/bindings_player_ffi.dart
```diff
@@ -375,14 +375,18 @@ class FlutterSoLoudFfi {
...
-  /// Return the handle of the sound, 0 if error
+  ///
+  /// Return the handle of the sound, [SoundHandle.error] if error
   SoundHandle play(
     SoundHash soundHash, {
     double volume = 1,
     double pan = 0,
     bool paused = false,
   }) {
     final handleId = _play(soundHash.hash, volume, pan, paused ? 1 : 0);
```
filiph, line 386: It would be great if the error was propagated somehow from C++ land, because otherwise we have to return an unknown error, which is always frustrating to devs.

alnitak: There can only be 2 kinds of errors in the `play()` C function:
- the given `soundHash` is not found
- the `Soloud::play` cannot play an already loaded sound returning an `UNKNOWN_ERROR`
So, I can return a `PlayerError.soundHashNotFound` or `PlayerError.unknownError`. For the latter, I don't know what the cause could be.

alnitak: here the [Soloud.play()](https://solhsa.com/soloud/basics.html) doc (search `Soloud.play()` in the page)

alnitak (approved):

PR #43, opened 2024-03-11
adding looping and start loop at position to play() and play3d()

alnitak:
## Description

Now `SoLoud.play()` and `SoLoud.play3d()` have also `looping` and `loopingStartAt` properties. They also can be set/get at any time with `SoLoud.getLoopPoint()` and `SoLoud.setLoopPoint()`.
Added `SoLoud.getLooping()` to get the looping state.
To test them I used the example app: in the 3D page the siren loop is played with `looping` property, in multi track page there is a check box and a slider to control these values.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/bindings_player_ffi.dart
+++ b/lib/src/bindings_player_ffi.dart
@@ -369,68 +369,50 @@ class FlutterSoLoudFfi {
   late final _getRelativePlaySpeed =
       _getRelativePlaySpeedPtr.asFunction<double Function(int)>();
 
-  /// Play already loaded sound identified by [soundHandle].
-  ///
-  /// [soundHandle] the unique sound hash of a sound
-  /// [volume] 1.0f full volume
-  /// [pan] 0.0f centered
-  /// [paused] 0 not pause
-  /// Return the handle of the sound, 0 if error
-  // SoundHandle play(
-  //   SoundHash soundHash, {
-  //   double volume = 1,
-  //   double pan = 0,
-  //   bool paused = false,
-  // }) {
-  //   final handleId = _play(soundHash.hash, volume, pan, paused ? 1 : 0);
-  //   return SoundHandle(handleId);
-  // }
-
-  // late final _playPtr = _lookup<
-  //     ffi.NativeFunction<
-  //         ffi.UnsignedInt Function(
-  //           ffi.UnsignedInt,
-  //           ffi.Float,
-  //           ffi.Float,
-  //           ffi.Int,
-  //         )>>('play');
-  // late final _play =
-  //     _playPtr.asFunction<int Function(int, double, double, int)>();
-
   /// Play already loaded sound identified by [hash]
   ///
   /// [hash] the unique sound hash of a sound
-  /// [volume] 1.0f full volume
-  /// [pan] 0.0f centered
-  /// [paused] 0 not paused
-  /// Return the error if any and a new [handle] of this sound
-  ({PlayerErrors error, SoundHandle handle}) play(
+  /// [volume] 1.0 full volume
+  /// [pan] 0.0 centered
+  /// [paused] false not paused
+  /// [looping] whether to start the sound in looping state.
+  /// [loopingStartAt] If looping is enabled, the loop point is, by default,
+  /// the start of the stream. The loop start point can be set with this
+  /// parameter, and current loop point can be queried with [getLoopingPoint]
+  /// and changed by [setLoopingPoint].
+  /// Return the error if any and a new [newHandle] of this sound
+  ({PlayerErrors error, SoundHandle newHandle}) play(
     SoundHash soundHash, {
     double volume = 1,
     double pan = 0,
     bool paused = false,
+    bool looping = false,
+    double loopingStartAt = 0,
   }) {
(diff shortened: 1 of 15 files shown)


lib/src/bindings_player_ffi.dart
```diff
@@ -369,37 +369,50 @@ class FlutterSoLoudFfi {
...
-  /// [soundHash] the unique sound hash of a sound
-  /// [volume] 1.0f full volume
-  /// [pan] 0.0f centered
-  /// [paused] 0 not pause
-  ///
-  /// Return the handle of the sound, [SoundHandle.error] if error
-  SoundHandle play(
+  /// Play already loaded sound identified by [hash]
+  ///
+  /// [hash] the unique sound hash of a sound
```
filiph, line 374: [nit] Should probably refer to `[soundHash]` as that's the name of the parameter.

lib/src/bindings_player_ffi.dart
```diff
@@ -451,6 +477,36 @@ class FlutterSoLoudFfi {
   );
   late final _setLooping = _setLoopingPtr.asFunction<void Function(int, int)>();
 
+  /// Get sound loop point value.
+  ///
+  /// [handle]
+  /// Returns the time in seconds.
+  double getLoopPoint(int handle) {
```
filiph, line 484: I'm confused. Shouldn't there be two looping points? The start and the end? Or is it only possible to construct a looping region that ends with the end of the whole sound?

lib/src/bindings_player_ffi.dart
```diff
@@ -451,6 +477,36 @@ class FlutterSoLoudFfi {
...
+      _lookup<ffi.NativeFunction<ffi.Double Function(ffi.UnsignedInt)>>(
+          'getLoopPoint');
+  late final _getLoopPoint =
+      _getLoopPointPtr.asFunction<double Function(int)>();
+
+  /// Set sound loop point value.
+  ///
+  /// [handle]
+  /// [time] in seconds.
+  void setLoopPoint(int handle, double time) {
```
filiph, line 498: This should take a `SoundHandle handle`, for consistency.

src/player.cpp
```diff
@@ -264,8 +266,15 @@ unsigned int Player::play(
         return 0;
 
     ActiveSound *sound = s->get();
-    SoLoud::handle newHandle = soloud.play(*sound->sound.get(), volume, pan, paused, 0);
-    sound->handle.emplace_back(newHandle);
+    SoLoud::handle newHandle = soloud.play(*sound->sound.get(), volume, pan, looping ? true : paused, 0);
```
filiph, line 269: > `looping ? true : paused`

Does this mean that setting `looping` automatically makes the sound paused?

UPDATE: Oh, I get it, you're starting it as paused because when it's looping, you want to be able to set up the looppoint etc. below. I'd add a comment explaining it, so as not confuse future code reader.

filiph (approved):
LGTM, though I'm a bit confused about the looping functionality (why it has only one loop point).

lib/src/bindings_player_ffi.dart
filiph: Ok, reading https://solhsa.com/soloud/attributes.html, and yeah, there's just one loop point. This seems weird to me but no reason to change this if the C++ part doesn't support regions.

lib/src/bindings_player_ffi.dart
alnitak: yep!

lib/src/bindings_player_ffi.dart
alnitak: right

src/player.cpp
alnitak: yes true, I write some explaination.
The problem that make me to do that, is the fact that, even if I set a `loopingStartAt`, the first time the sound plays it plays from the beginning. From the 2nd time the starting point is used

lib/src/bindings_player_ffi.dart
alnitak: I thought there was an ending point too, but there isn't. I think there is a way to do that, I will look at it later on, meanwhile it's something that could be useful. It is mentioned also [here](https://github.com/jarikomppa/soloud/issues/15#issuecomment-375354894).

My wish to have an endpoint, was to gain something similar to `sprite sheets` used in games, where a sequence of sprites are all in the same image. Likewise, an audio could have a bunch of different sounds in the same audio file.

AyoAlfonso:
Hi ! I am using your package manipulate a track, and it seems I can only set the start of the loop and not the end of the loop on a track, is there a technical reaspon for this? 

@alnitak

alnitak:
Hi @AyoAlfonso,

please, we can more easily track this if you file an issue for a feature request.

AyoAlfonso:
I don't think I am asking for a feature request, I am asking if it even exists at all

alnitak:
Well, it doesn't exist now because SoLoud lib doesn't have it, but it can be planned and implemented in the future.

PR #47, opened 2024-03-13
Throw exceptions instead of returning `PlayerError`

filiph:
## Description

- `SoLoud` methods now throw instead of returning a `PlayerError` object.
  This is a massive breaking change, but it makes the package API
  more idiomatic and easier to use.
  
  Before:

  ```dart
  final ret = await SoLoud.play(sound);
  if (ret.error != PlayerError.noError) {
    print('Oh no! ${ret.error}');
  } else {
    print('Playing sound with new handle: ${ret.newHandle}');
  }
  ```

  After:

  ```dart
  try {
    final handle = await SoLoud.play(sound);
    print('Playing sound with new handle: $handle');
  } on SoLoudException catch (e) {
    print('Oh no! $e');
  }
  ```

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


filiph:
I haven't yet documented everything and there might be exceptions that we never throw. But the code compiles, examples and tests work.

filiph:
Btw, `PlayerErrors` is still used, but only internally. In public APIs, they are thrown as exceptions.

I haven't touched `CaptureErrors` because that's a whole other massive piece of work, and I'd rather not spend that kind of time on capture at this point.

alnitak:
Agree, we will think about capture later.
L(very)GTM!

filiph:
I've completed this transition (by splitting errors into those that come from the C++ side, and those that come from the Dart side. I also added a unit test that verifies that the PlayerErrors enum is consistent with its usage (it should be much harder now to accidentally introduce inconsistencies). Run by `flutter test`.

I know there's a bunch of documentation work still to be done, but I'm postponing it for later.

If you're happy with this, feel free to merge and publish as another preview version.

alnitak (approved):
Great job!

Issue #52, opened 2024-03-15
proposal: Another set of renames


filiph:
**Description**

I have another set of renames that I'd like to run by you (when you have time!). Basically, I'm reading the SoLoud page now (e.g. https://solhsa.com/soloud/concepts.html) and I think we might want to stick to their terminology. 

![](https://solhsa.com/soloud/images/concepts.png)

For example, `SoundProps` might become `AudioSource`. Possibly `SoLoud.initialize()` to `SoLoud.init()`, and `shutdown()` to `deinit()`, although here I'm uncertain since those are against Dart style. But looking at the C# wrapper, for example, I can see that they also break C# conventions (https://solhsa.com/soloud/cs_api.html). Or maybe `deinit()` should be a synchronous "kill switch" for when there's no time at the very end of the app/game's lifecycle — and we keep `shutdown()` for a softer, more thorough exit.

Apart from renaming, I'd love to move the documentation as close to the original (C++) library as possible.

filiph:
Possibly `SoundHandle` should be `Voice`? Although even the C++ docs are inconsistent and call it mostly "sound handle".

filiph:
cc @alnitak for you thoughts on this

alnitak:
> Or maybe deinit() should be a synchronous "kill switch"

I just came across this situation with #54 ! :) So I also think your `deinit()` "kill switch" is a good idea.

> For example, `SoundProps` might become `AudioSource`.

IIRC I named `SoundProps` (SoundProperties shortened), to store the ways to manage it in [native code](https://github.com/alnitak/flutter_soloud/blob/e79a897a655ad6ac61d23b9b9d50b2e72b8d9c1e/src/player.h#L31).
Into `SoundProps` we could add for example the audio file name, like in native, but also any other properties we like. It's not  really like [`Soloud::AudioSource`](https://solhsa.com/soloud/audiosource.html) which in `SoundProps` could be represented by `SoundHash`.
Anyway, I like `AudioSource` name more than `SoundProps`!

> Possibly SoundHandle should be Voice? Although even the C++ docs are inconsistent and call it mostly "sound handle".

I recall that I get a bit confused too when I read `voice`. Maybe it's a deeper Soloud lib concept needed maybe if/when we will go deeper integrating the lib into the plugin.  The Soloud author says "The handle is an integer, and internally tracks the voice and sound id, as well as an "uniqueness" value.", so I think the "voice" could be thought as a sub-type of a handle.

[...]

PR #56, opened 2024-03-17
added synchronous deinit()

alnitak:
## Description

- added `deinit()` which shutdown the player synchronously #52

Please feel free to check the code, I'm not sure I've covered all the steps needed to shut down.
I also did a test that initializes and deinit the player with an interval that gradually gets shorter until it reaches 0 and the checks if the sound is disposed correctly.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


example/tests/tests.dart
```diff
@@ -29,28 +31,36 @@ void main() async {
...
+  //   },
+  // );
+
+  // await runZonedGuarded(
+  //   () async => test2(),
+  //   (error, stack) {
+  //     stderr.writeln('TEST error: $error\nstack: $stack');
+  //     exitCode = 1;
+  //   },
+  // );
```
filiph, line 63: Is this supposed to be commented out still?

lib/src/soloud.dart
```diff
@@ -628,6 +629,38 @@ interface class SoLoud {
...
+
+    /// if not already initialized, just return.
+    if (!_isInitialized) {
+      // The engine isn't initialized.
+      _log.warning('deinit() called when the engine is not initialized');
+      return;
+    }
+
+    /// reset broadcast and kill isolate
+    _cleanUpUnsuccessfulInitialization();
```
filiph, line 657: I'd rather spell this out (inline). That method is really meant only to clean up after unsuccessful initialization.

lib/src/soloud.dart
```diff
@@ -628,6 +629,38 @@ interface class SoLoud {
...
+    _isInitialized = false;
```
filiph, line 658: Are we not calling `SoLoud::deinit()`?

alnitak:
Previously, `shutdown()` was a Future because `stopIsolate()` and `stopEngine()` were two separate methods. Now, however, might we consider using this instead of `shutdown()` if the new synchronous `deinit()` works?

lib/src/soloud.dart
```diff
@@ -628,6 +629,46 @@ interface class SoLoud {
...
+    _mainToIsolateStream = null;
+    _returnedEvent?.close();
+    _returnedEvent = null;
+    _isolate?.kill();
+    _isolate = null;
+    _isEngineInitialized = false;
+    _isInitialized = false;
+
+    SoLoudController().soLoudFFI.disposeAllSound();
+    SoLoudController().soLoudFFI.dispose();
```
filiph, line 668: Just to be clear, this actually calls [`deinit()` on the C++ side](https://solhsa.com/soloud/basics.html), correct?

alnitak: Correct! An FFI method was missing in `bindings_player_ffi.dart` to check this. I double-checked that the player is really disposed even after 0ms since the end of `initialized()`. So we can think about the [previous comment](https://github.com/alnitak/flutter_soloud/pull/56#issuecomment-2003774773).

filiph (approved):
One outstanding question/suggestion, otherwise LGTM!

filiph:
So, we do still need async `shutdown()` for "proper" shutdown (including things like cleaning up temp directory etc.), but there's `deinit()` which is a kill switch. Just writing it down in case I'm wrong about this so you can correct me.

filiph:
> So, we do still need async shutdown() for "proper" shutdown (including things like cleaning up temp directory etc.), but there's deinit() which is a kill switch.

Actually, I just remember we don't do any cleaning up of the temp directory on shutdown. If we can just synchronously shutdown, I vote for doing that, and having only `void deinit()` (which would take over the role of the async `shutdown()`).

alnitak:
> Actually, I just remember we don't do any cleaning up of the temp directory on shutdown. If we can just synchronously shutdown, I vote for doing that, and having only `void deinit()` (which would take over the role of the async `shutdown()`).

Ok, we will do that in another PR then.

PR #57, opened 2024-03-18
Rename SoundProps to AudioSource

### LINKED ISSUE #52: proposal: Another set of renames

filiph:
## Description

- Renamed `SoundProps` to `AudioSource`. Quick fix available.
- Added `SoundProps` as a temporary alias of `AudioSource`, for easier migration.

Fixes #52.

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ to `2.0.0-pre.2` and beyond.
 - Renamed `SoLoud.setFxParams` to `SoLoud.setFilterParameter`. 
   This mimics the C++ API name.
   Quick fix available.
+- Renamed `SoundProps` to `AudioSource`. Quick fix available.
 
 #### 2.0.0-pre.1 (12 Mar 2024)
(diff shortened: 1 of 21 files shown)


alnitak (approved):
LGTM!

PR #60, opened 2024-03-19
fix: asynchronous initialize-deinit bug

alnitak:
## Description

Fix for #56
Doing some more testing on the `deinit()`, the bug was seen in `test4()` when trying to use unawaited `initialize()` and suddenly `deinit()`. Not happening with `await initialize()`. This caused a failed test.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/example/tests/tests.dart
+++ b/example/tests/tests.dart
@@ -37,29 +37,29 @@ void main() async {
     },
   );
 
-  // await runZonedGuarded(
-  //   () async => test3(),
-  //   (error, stack) {
-  //     stderr.writeln('TEST error: $error\nstack: $stack');
-  //     exitCode = 1;
-  //   },
-  // );
+  await runZonedGuarded(
+    () async => test3(),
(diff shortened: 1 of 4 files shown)


alnitak (commented):
Instead of removing `shutdown()` we may think about leaving it and renaming `deinit()` to `shutdownSync()`? What do you think?

alnitak:
> My question is whether we even need shutdown(). If deinit() calls SoLoud::deinit() and just kills the isolate, maybe there is no reason to wait for anything?

I think that `deinit()` does its job without waiting, just to be sure you also approved this change.

filiph:
In that case, I think we can just remove `shutdown()` altogether?

filiph:
(I can do that in a follow up PR.)

alnitak:
No problem, I'll do in this PR

PR #66, opened 2024-03-22
suppress cpp warnings

alnitak:
## Description

Suppress all warnings coming from the SoLoud lib.
On macOS and iOS all warnings are suppressed.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -42,7 +42,7 @@
             "type": "cppvsdbg",
             "request": "launch",
             "args": [],
-            "program": "${workspaceFolder}/example/build/windows/runner/Debug/flutter_soloud_example.exe",
+            "program": "${workspaceFolder}/example/build/windows/x64/runner/Debug/flutter_soloud_example.exe",
             "cwd": "${workspaceFolder}"
         },
         {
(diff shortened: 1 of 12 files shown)


alnitak:
@filiph, please can you try compiling from a clean project and see if there are still warnings? The only solution I found for macOS and iOS is to suppress all warnings. I didn't find how to suppress warnings for only the `src/soloud` dir.

filiph:
To be honest, I'm not sure about this. Are we sure we want to inhibit all warnings? That sounds dangerous. Especially for the package's own development — imagine not seeing a valid warning after we do some change to the C++ code, and introducing some subtle bug this way.

In my opinion, there are a few ways to address the issue:

- Leave the warnings be. Just document that they're there, and that if they come from `src/soloud`, they're probably nothing to worry about. Complexity: minimal.
- Try to actually address the warnings, changing the code in `src/soloud`. Complexity: possibly huge.
- Somehow suppress the warnings, but _not_ for `flutter_soloud` package developers. In other words, you and me and other volunteers can see them when working on the package, but regular users don't. Complexity: ???

alnitak:
Many of the warnings coming from SoLoud lib are really inifluent if not wrong catches. All those from the SoLoud lib can be suppressed safely.
I agree with you that warnings are mandatory when developing.

> Somehow suppress the warnings, but not for flutter_soloud package developers. In other words, you and me and other volunteers can see them when working on the package, but regular users don't. Complexity: ???

This could be the way. Maybe using an environment variable in the IDE running configuration. I write down this in my TODOs and think about this after we release v2.0.
Since the few changes to this PR will not be useful for the above solution, I'll close this. 
Thanks!

PR #73, opened 2024-04-04
v2.0

alnitak:
## Description

- removed deprecated
- CHANGELOG updated for v2.0

**edit**
- removed `MessageEvents.disposeEngine` and `MessageEvents.stopLoop` enums alongside `_stopLoop()`  and `_disposeEngine()` methods because were used in the old async `disposeEngine()`. All that work is now done by `deinit()`.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [x] 🗑️ Chore


alnitak (commented):
Do I need to delete the `fix_data.yaml` file and the `test_fixes` directory?

In the CHANGELOG I have written the major changes since v1.2.5. Is that enough? Please, take a look if I missed something. I left the `pre` versions for reference before deleting them (if you agree to delete them).

Tests passed on Linux, Android and Windows.

filiph (approved):
LGTM! Let me play around with it a bit and then we can publish on pub.dev.

Thanks!

filiph:
> Do I need to delete the fix_data.yaml file and the test_fixes directory?

No no, let's keep these. I don't know if they'll still work but it's worth having them around for a while.

> In the CHANGELOG I have written the major changes since v1.2.5. Is that enough? Please, take a look if I missed something. 

Makes sense to me!

> I left the pre versions for reference before deleting them (if you agree to delete them).

I think we can keep those. The people who're still on 1.x will be interested in seeing the path, I think.

> Tests passed on Linux, Android and Windows.

I'll try macOS and iOS once it lands.

Issue #81, opened 2024-04-27
fix: Init error on hot restart


daniloapr:
**Description**

I'm instantiating the `SoLoud.instance` in a singleton class and calling `init()` when the singleton's constructor. When I hot reload the application, it throws a `SoLoudPlayerAlreadyInitializedException` (logs below). However, when I try to load assets, it throws the exception `SoLoudNotInitializedException`. This is preventing me from using the library.

**Steps To Reproduce**

1. Create a singleton class and call `SoLoud.instance.init()`
2. Hot restart the app

**Expected Behavior**

The previous instance should be refreshed and no init() error should occur

**Logs**

```
flutter: FINEST: 2024-04-27 12:22:15.502315: initialize() called
flutter: INFO: 2024-04-27 12:22:15.526642: Temporary directory initialized at /var/mobile/Containers/Data/Application/843B8881-FF5A-4429-AE75-B41CF54866E8/Library/Caches/SoLoudLoader-Temp-Files
flutter: FINEST: 2024-04-27 12:22:15.694999: _initEngine() called
flutter: FINEST: 2024-04-27 12:22:15.704450: main isolate received: {event: MessageEvents.initEngine, args: (), return: PlayerErrors.playerAlreadyInited (The player has already been inited!)}
flutter: SEVERE: 2024-04-27 12:22:15.706821: _initEngine() result: PlayerErrors.playerAlreadyInited (The player has already been inited!)
flutter: SEVERE: 2024-04-27 12:22:15.709483: _initEngine() failed with error: PlayerErrors.playerAlreadyInited (The player has already been inited!)
```

```
SoLoudNotInitializedException: SoLoud has not been initialized yet. Call `SoLoud.initialize()` first and await it, or await the `SoLoud.initialized` Future elsewhere. Alternately, you can check the synchronous `SoLoud.isInitialized` bool just before calling any SoLoud method
```

**Code**
```dart
// on app initialization
GetIt.instance.registerSingleton<SoLoudImpl>(SoLoudImpl());
```

```dart
import 'package:flutter/foundation.dart';
import 'package:flutter_soloud/flutter_soloud.dart';
import 'package:sentry_flutter/sentry_flutter.dart';

class SoLoudImpl {
  final player = SoLoud.instance;

  SoLoudImpl() {
    try {
      player.init();
    } catch (e) {
      debugPrint(e.toString());
      Sentry.captureException(e);
    }
  }

  AudioSource? beep;

  /// Load assets
[...]

PR #82, opened 2024-05-06
fix: Init error on hot restart #81

alnitak:
## Description

Fix hot reload `SoLoudPlayerAlreadyInitializedException` issue #81.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/soloud.dart
+++ b/lib/src/soloud.dart
@@ -292,7 +292,11 @@ interface class SoLoud {
     // the release of the native player.
     // Just deinit the engine to be re-inited later.
     if (SoLoudController().soLoudFFI.isInited()) {
-      _log.warning("init() called when the native player is already initialized. This is expected after a hot restart but not otherwise. If you see this in production logs, there's probably a bug in your code. You may have neglected to deinit() SoLoud during the current lifetime of the app.");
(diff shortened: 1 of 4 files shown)


lib/src/soloud.dart
```diff
@@ -287,6 +287,14 @@ interface class SoLoud {
       return;
     }
 
+    // if `!_isInitialized` but the engine is initialized in native, therefore
+    // the developer may have carried out a hot reload which does not imply
+    // the release of the native player.
+    // Just deinit the engine to be re-inited later.
+    if (SoLoudController().soLoudFFI.isInited()) {
+      deinit();
```
filiph, line 295: I think this calls for a warning in logs.

```suggestion
      _log.warning("init() called when the native player is already initialized. This is expected after a hot restart but not otherwise. If you see this in production logs, there's probably a bug in your code. You may have neglected to deinit() SoLoud during the current lifetime of the app.");
      deinit();
```

I'm writing this in github web interface so the formatting is wrong but you get the idea.

filiph (approved):
LGTM with one suggestion.

alnitak:
Great! Thank you.

Issue #83, opened 2024-05-21
fix: handles become invalid at unknown points 


LeeMatthewHiggins:
If I play a sound and hold on to the handle to set value later, if the sound is looping everything is fine.
If not when I try and set volume using that handle I get a crash.
"The player is not initialized (on the C++ side)."

**Steps To Reproduce**

1. Play a non looping short sound, hold on to the handle
2. Call setVolume using the handle later after the sound has finsihed playing
3. Crash (The player is not initialized (on the C++ side).)
*line 1236 in soloud.dart*

**Expected Behavior**

The system should not crash, handles should remain valid
OR
We need a system to tell us when handles are invalid

**Additional Context**

iOS, flutter_soloud: ^2.0.1

alnitak:
Hi, thanks for your feedback.

Are you sure you got a crash and not an exception? The  `setVolume()` can throw `SoLoudNotInitializedException` and for your use-case, it would be better to put it inside a `try-catch`.
Also, you are right, the error message is wrong, I'll fix it soon. It should print out something related to the invalid handle.

To know when a handle is no longer valid, you can listen to the [`AudioSource`](https://pub.dev/documentation/flutter_soloud/latest/flutter_soloud/AudioSource-class.html) you got from `loadFile`. It has `soundEvents` stream to listen to which has the [`event.handleIsNoMoreValid`](https://pub.dev/documentation/flutter_soloud/latest/flutter_soloud/SoundEventType.html)  event you are looking for.

alnitak:
The sound handles are the representation of the loaded sound and are used internally to control them. 

For example, you can play an unlimited number of the same instance of a sound and every instance has its unique handle ID. By `control them` I mean that you can play, pause, get info, or set a filter on that playing handle independently from each other.

So a sound (the `AudioSource`) can have multiple active handles. They are active when playing or even when paused, but when they reach the end or when they are stopped, the handle becomes invalid and its unique ID is removed from the internal handles list.

The [`setVolume()`](https://pub.dev/documentation/flutter_soloud/latest/flutter_soloud/SoLoud/setVolume.html) you are using acts on a specific instance (handle) of a sound. Maybe could be better for your use-case to use [`setGlobalVolume()`](https://pub.dev/documentation/flutter_soloud/latest/flutter_soloud/SoLoud/setGlobalVolume.html) instead?

LeeMatthewHiggins:
[...]

PR #85, opened 2024-05-23
removed cpp exception check from setVolume

alnitak:
## Description

Since only `setVolume()` was checking for a possible "handle not found" exception, I've removed that.

All the other methods that need a `SoundHandle` as a parameter don't throw a cpp exception and don't log anything. Therefore, when using a voice handle that is already ended/stopped nothing happens as discussed in issue #83.

We could think about just displaying a log for all the methods that need a `SoundHandle` without throwing.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/lib/src/exceptions/exceptions.dart
+++ b/lib/src/exceptions/exceptions.dart
@@ -100,7 +100,7 @@ abstract class SoLoudCppException extends SoLoudException {
       case PlayerErrors.playerAlreadyInited:
         return const SoLoudPlayerAlreadyInitializedException();
       case PlayerErrors.soundHandleNotFound:
-        return const SoLoudSoundHashNotFoundCppException();
+        return const SoLoudSoundHandleNotFoundCppException();
     }
   }
 }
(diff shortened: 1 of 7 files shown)


lib/src/exceptions/exceptions.dart
```diff
@@ -99,6 +99,8 @@ abstract class SoLoudCppException extends SoLoudException {
         return const SoLoudFilterAlreadyAddedException();
       case PlayerErrors.playerAlreadyInited:
         return const SoLoudPlayerAlreadyInitializedException();
+      case PlayerErrors.soundHandleNotFound:
+        return const SoLoudSoundHashNotFoundCppException();
```
filiph, line 103: It's confusing to have "soundhandle" error create a "soundhash" exception. We should pick one.

filiph (approved):
LGTM with one comment.

lib/src/exceptions/exceptions.dart
alnitak: Ouch, my bad. Thank you!

PR #91, opened 2024-06-04
added getPan, setPan and setPanAbsolute

### LINKED ISSUE #90: feat: Add functions to directly get and set pan
**Description**

As of the time of writing, there is no getPan or setPan functions to directly set and get the pan value of a handle. However, soloud does provide such functions. The only way to set pan as of yet, is to provide a `pan` value to `play()`, or use the `fadePan` method, but that is not appropriate when you just need to instantly change `pan`, or when you need to just get the current value.

**Requirements**

- [ ] Add a function to get the current pan value of a handle.
- [ ] Add a function to set the current pan value of a handle.

**Additional Context**

We are making a game using this, and directly getting / setting pan is required for us, so we would appreciate it if this can be looked into.
Thanks.

alnitak:
## Description

- added `getPan()`, `setPan()` and `setPanAbsolute()`  to fix #90.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/bindings_player_ffi.dart
+++ b/lib/src/bindings_player_ffi.dart
@@ -739,6 +739,55 @@ class FlutterSoLoudFfi {
       'setVolume');
   late final _setVolume = _setVolumePtr.asFunction<int Function(int, double)>();
 
+  /// Get a sound's current pan setting.
+  ///
+  /// [handle] the sound handle.
+  /// Returns the range of the pan values is -1 to 1, where -1 is left, 0 is
+  /// middle and and 1 is right.
+  double getPan(int handle) {
+    // Note that because of the float<=>double conversion precision error
+    // (SoLoud lib uses floats), the returned value is not precise. Here we set
+    // a rounding of 5 digits
+    final ret = (_getPan(handle) * 100000).toInt() / 100000;
+    return ret;
+  }
+
+  late final _getPanPtr =
+      _lookup<ffi.NativeFunction<ffi.Double Function(ffi.UnsignedInt)>>(
+          'getPan');
+  late final _getPan = _getPanPtr.asFunction<double Function(int)>();
+
+  /// Set a sound's current pan setting.
+  ///
+  /// [handle] the sound handle.
+  /// [pan] the range of the pan values is -1 to 1, where -1 is left, 0 is
+  /// middle and and 1 is right.
+  void setPan(int handle, double pan) {
+    return _setPan(handle, pan);
+  }
+
+  late final _setPanPtr = _lookup<
+          ffi.NativeFunction<ffi.Void Function(ffi.UnsignedInt, ffi.Double)>>(
+      'setPan');
+  late final _setPan = _setPanPtr.asFunction<void Function(int, double)>();
+
+  /// Set the left/right volumes directly.
+  /// Note that this does not affect the value returned by getPan.
+  ///
+  /// [handle] the sound handle.
+  /// [panLeft] value for the left pan.
+  /// [panRight] value for the right pan.
+  void setPanAbsolute(int handle, double panLeft, double panRight) {
+    return _setPanAbsolute(handle, panLeft, panRight);
+  }
+
+  late final _setPanAbsolutePtr = _lookup<
+      ffi.NativeFunction<
+          ffi.Void Function(
+              ffi.UnsignedInt, ffi.Double, ffi.Double)>>('setPanAbsolute');
+  late final _setPanAbsolute =
+      _setPanAbsolutePtr.asFunction<void Function(int, double, double)>();
+
   /// Check if a handle is still valid.
   ///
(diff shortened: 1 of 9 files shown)


lib/src/bindings_player_ffi.dart
```diff
@@ -739,6 +739,55 @@ class FlutterSoLoudFfi {
...
+  /// Get a sound's current pan setting.
+  ///
+  /// [handle] the sound handle.
+  /// Returns the range of the pan values is -1 to 1, where -1 is left, 0 is
+  /// middle and and 1 is right.
+  double getPan(int handle) {
+    // Note that because of the float<=>double conversion precision error
+    // (SoLoud lib uses floats), the returned value is not precise. Here we set
+    // a rounding of 5 digits
+    final ret = (_getPan(handle) * 100000).toInt() / 100000;
```
filiph, line 751: Frankly, I don't understand why we need to do this rounding. Is it not enough to just return the double?

lib/src/soloud.dart
```diff
@@ -1232,6 +1232,53 @@ interface class SoLoud {
...
+  }
+
+  /// Set a sound's current pan setting.
+  ///
+  /// [handle] the sound handle.
+  /// [pan] the range of the pan values is -1 to 1, where -1 is left, 0 is
+  /// middle and and 1 is right.
+  ///
+  /// Throws [SoLoudNotInitializedException] if the engine is not initialized.
+  void setPan(SoundHandle handle, double pan) {
```
filiph, line 1256: What happens if the user tries to set pan to `123.0` or `double.infinity`? Should we `assert(-1 <= pan && pan <= 1)` and then `.clamp()` the value?

src/player.cpp
```diff
@@ -443,6 +443,30 @@ void Player::setVolume(SoLoud::handle handle, float volume)
...
+{
+    return soloud.getPan(handle);
+}
+
+void Player::setPan(SoLoud::handle handle, float pan)
+{
+    if (pan > 1.0f)
+        pan = 1.0f;
+    if (pan < -1.0f)
+        pan = -1.0f;
```
filiph, line 456: Oh, I see that we're sanitizing the input here in C code. Since this is likely not a performance bottleneck, I'd rather sanitize it in Dart land, since that's closer to the user's debugger.

src/player.cpp
```diff
@@ -443,6 +443,30 @@ void Player::setVolume(SoLoud::handle handle, float volume)
...
+    soloud.setPan(handle, pan);
+}
+
+void Player::setPanAbsolute(SoLoud::handle handle, float panLeft, float panRight)
+{
+    if (panLeft > 1.0f) panLeft = 1.0f;
+    if (panLeft < -1.0f) panLeft = -1.0f;
+    if (panRight > 1.0f) panRight = 1.0f;
+    if (panRight < -1.0f) panRight = -1.0f;
+    soloud.setPanAbsolute(handle, panLeft, panRight);
```
filiph, line 466: same here for panAbsolute: ideally sanitize (and assert) in Dart-land

filiph (approved):
Looks good! I have some nit comments.

filiph:
> can you check if it grabs the state changes on iOS? For example when you put the app in the background or when the app is back in the foreground or maybe when a call is incoming while flutter_soloud is playing something, you should see a Log.fine() message (Audio engine state changed: [event]). This is out of scope of the PR, but I have added it while adding the others for simplicity. Thank you!

I don't see the change here. Is this something that landed in another PR? 

I should have my iOS testing setup ready in a few minutes. Can test.

src/player.cpp
alnitak: You are right! Better to let the dev know what he is trying to do :)

lib/src/bindings_player_ffi.dart
filiph: FWIW, if this is about testing, I'd just document this in the function signature ("Note that if you `setPan()` to `0.8` and then `getPan()`, you might get a slightly different number, such as `0.800000042353`. This is expected since the internal audio engine uses `float` instead of `double`, and so there are rounding errors.").

Then, in unit tests, just use the `closeTo()` operand to deal with the tiny changes.

alnitak: Doing so we add something that the dev should remember when seeing a strange number given back. If you think this is the way, no problem to fix it.

Saying that I realized that this happens also for other functions like `setVolume()` - `getVolume()`! So, that ok :)

filiph: Yeah, I think it's great to be aware of this and document this, but I don't think it's our place to quantize values for the developer in order to make the values "look right". I totally understand the sentiment, of course. I just think it's adding a functionality people might not necessarily want that needlessly adds more things to the package, too.

PR #103, opened 2024-07-08
Web

alnitak:
## Description

The web platform is now supported.

In the `web` directory, there is a `compile_wasm.sh` script that generates the `.js` and `.wasm` files for the native C code located in the `srd` dir. Run it after installing *emscripten*. There is also a `compile_web.sh` to compile the web worker needed by native code to communicate with Dart. The generated files are already provided, but if it is needed to modify C/C++ code or the `web/worker.dart` code, the scripts must be run to reflect the changes.

The `compile_wasm.sh` script uses the `-O3` code optimization flag.

To see a better errors logs, use `-O0 -g -s ASSERTIONS=1` in `compile_wasm.sh` and rerun the script.

To add the plugin to a web app, add the following line to the `<body>` section of `web/index.html`:
`<script src="assets/packages/flutter_soloud/web/libflutter_soloud_plugin.js" defer></script>`

## The problems

The AudioIsolate was causing many problems related to web workers when trying to port to web. It was used to monitor all the sound states and to send some operations to the native code. These operations were not working inside a JS Worker because web audio is not supported.

The `AudioIsolate` [has been removed](https://github.com/alnitak/flutter_soloud/pull/89) and all the logic has been implemented natively. Events like `voice ended` are sent from C back to Dart. However, since it is not possible to call Dart from a native thread (the audio thread), a new web worker is created using the WASM `EM_ASM` directive. This allows sending the `audio finished` event back to Dart via the worker. Here an example just to let me explain and receive suggestions/critiques:
```CPP
// On C side define an inline function to create the Web Worker and another
// to send messages to the listening worker in Dart.

void createWorkerInWasm()
{
	EM_ASM({
		...
		// Create a new Worker
		var workerUri = "assets/packages/flutter_soloud/web/worker.dart.js";
		Module.wasmWorker = new Worker(workerUri);
	});
}

void sendToWorker(const char *message, int value)
{
	EM_ASM({
			// Send the message
			Module.wasmWorker.postMessage(JSON.stringify({
				"message" : UTF8ToString($0),
				"value" : $1
			}));
		}, message, value);
}

```

```Dart
// On Dart side listen to the messages.
@JS('Module.wasmWorker')
external web.Worker wasmWorker;

void setDartEventCallbacks() {
	web.Worker _worker = wasmWorker;
	_worker?.onmessage = ((web.MessageEvent event) {
      _outputController?.add(event.data.dartify());
    }).toJS;
}
```

The same problem happens using `dart:ffi`, it is also not possible to call a function directly from a native thread (audio thread) to Dart without using send. Here, `NativeCallable` helps. With `NativeCallable`, it's not necessary to import sendPort and receivePort into the native code, which are part of ffi and thus not compatible with the web.

## Notes

Acquiring audio data is (was?) experimental, the following methods are now deprecated and this functionality is now in the `AudioData` class.
- `@experimental SoLoud.getAudioTexture2D()`
- `@experimental SoLoudCapture.getCaptureAudioTexture2D()`

It is not possible to read a local audio file directly on the web. For this reason, `loadMem()` has been added, which requires the `Uint8List` byte buffer of the audio file.
**IMPORTANT**: `loadMem()` with mode `LoadMode.memory` used the on web platform will freezy the UI for the time needed to decompress the audio file. Please use it with mode `LoadMode.disk` or load your sound when the app starts.

In addition to acquire data as a 2D texture (`getAudioTexture2D`), with `AudioData` class is now possible to acquire audio as `linear`, which represents the FFT+wave vector, or just the `wave` data vector for better performance. With this class, it is also possible to choose to acquire data from the player or from the microphone.

**`loadUrl()`** produce the following error when the app is run:
>> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.learningcontainer.com/wp-content/uploads/2020/02/Kalimba.mp3. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.

This is due for the default beavior of http servers which don't allow to make requests outside their domain. Refer [here](https://enable-cors.org/server.html) to learn how to enable your server to handle this situation.
Instead, if you run the app locally, you could run the app with something like the following command:
`flutter run -d chrome --web-renderer canvaskit --web-browser-flag '--disable-web-security' -t lib/main.dart --release`

**test/test.dart** has been changed and a GUI has been added to simplify the logs. It was using `dart:io` which is not compatible with the web, also `stdout()`/`stderr` were not output logs to the Android console.

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation
- [x] 🗑️ Chore


example/tests/tests.dart
```diff
@@ -1,100 +1,258 @@
 import 'dart:async';
-import 'dart:io';
+import 'dart:ui';
```
alnitak, line 2: Feel free to revert this. The problem in the previous was that `dart:io` is not compatible with web. Also the `std*()` didn't output correctly in the Android console.

lib/src/bindings/audio_data_web.dart
```diff
@@ -0,0 +1,75 @@
...
+      SoLoudController().captureFFI.getCaptureWave;
+
+  final CaptureErrors Function(AudioData) captureTexture2DCallback =
+      SoLoudController().captureFFI.getCaptureAudioTexture2D;
+
+  final void Function(AudioData) captureAudioTextureCallback =
+      SoLoudController().captureFFI.getCaptureAudioTexture;
+
+  void allocSamples() {
+    _samplesPtr = wasmMalloc(512 * 256 * 4);
```
alnitak, line 32: Here I allocate the max amount of bytes needed by `*texture2D*()` and used that memory for all kind of audio samples we want to acquire. I found not convenient to allocate the memory depending of the type we want to acquire (ie the `getAudioTexture()` just needs 256*4 bytes).

lib/src/bindings/js_extension.dart
```diff
@@ -0,0 +1,438 @@
+// ignore_for_file: public_member_api_docs
+
+import 'dart:js_interop';
```
alnitak, line 3: Here I don't know well if I choosen the right solution to interop with JS. All the functions compiled to WASM are exposed globally. I think there is a way to put them inside an `extension type` but I didn't find a solution.

lib/src/soloud.dart
```diff
@@ -202,6 +203,7 @@ interface class SoLoud {
   /// unnecessary, as the amount of data will be finite.
   /// The default is `false`.
   Future<void> init({
+    // TODO(filip): remove deprecation?
```
alnitak, line 206: we forgot to take a look at this deprecation.

lib/src/soloud.dart
```diff
@@ -242,7 +246,6 @@ interface class SoLoud {
       // Initialize [SoLoudLoader]
       _loader.automaticCleanup = automaticCleanup;
 
-      // TODO(filip): The Loader is not compatible with web!
       await _loader.initialize();
```
alnitak, line 249: the `Loader` has been splitted into `dart:io` and `dart:html` conditional imports.

lib/src/sound_hash.dart
```diff
@@ -37,7 +37,8 @@ extension type SoundHash._(int hash) {
   @internal
   factory SoundHash.random() {
     // Dart must support 32 bit systems.
-    const largest32BitInt = 1 << 32;
+    // Shifting by 31 because the leftmost bit is used for the sign.
+    const largest32BitInt = 1 << 31;
```
alnitak, line 41: This was only visible on web because JS complained about a number too big :)

pubspec.yaml
```diff
@@ -57,3 +58,8 @@ flutter:
         ffiPlugin: true
       windows:
         ffiPlugin: true
+
+  assets:
```
alnitak, line 62: Here I am not sure what to do. These assets are only needed on the web. Is there a way to do not include them on other platforms?

filiph:
Thanks, @alnitak, this is excellent! I'm going to have a look this evening or tomorrow. I love the deep context that you've provided.

I also asked the community (on Mastodon and Twitter) to have a look because, well, the more the merrier in this case. I could easily overlook something, and I'm also a complete beginner when it comes to WASM or emscripten.

filiph:
> IMPORTANT: loadMem() with mode LoadMode.memory used the on web platform will freezy the UI for the time needed to decompress the audio file. Please use it with mode LoadMode.disk or load your sound when the app starts.

This is clearly not something we'll address in this PR, but I wonder why this happens. I thought the audio decompression happens in C, and therefore on a non-UI thread, correct?

lib/src/bindings/audio_data_web.dart
filiph: I think this should be commented in the code. If I saw `512 * 256 * 4` without any other context, I wouldn't know what this means, and would be afraid to touch it.

Also, as an aside, I suggest we find ways to remove things from the package's API surface. There's a lot of work here that you need to do just in order to maintain `AudioData`, and I don't think it's a feature that many people use. Of course, as soon as you have a feature, _some_ people will start using it, and then it's really hard to remove it because you don't want to disappoint / anger folks.

lib/src/soloud.dart
filiph: Thanks for pointing it out! I'll have a look in a different PR, this one is already large.

pubspec.yaml
filiph: I'm not sure, but I think for that we would need a federated plugin (https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins). That's quite a lot of additional work, though. 

(There are open issues on conditional bundling of assets https://github.com/flutter/flutter/issues/65065 and platform specific assets https://github.com/flutter/flutter/issues/8230. I suggest we put a big 👍 on those issues.)

For now I think it's okay to leave everything here? We can tally the size of the additional files and ask people: is web support worth the additional X megabytes on all other platforms, or do you want to wait for us to do it "properly" first?

lib/src/soloud.dart
```diff
@@ -998,17 +1024,11 @@ interface class SoLoud {
   /// and `1.0` meaning full volume.
   ///
   /// Throws [SoLoudNotInitializedException] if the engine is not initialized.
-  ///
-  /// Note that if you `setGlobalVolume()` to `0.8` and then
-  /// `getGlobalVolume()`, you might get a slightly different number,
-  /// such as `0.800000042353`.
-  /// This is expected since the internal audio engine uses float
-  /// instead of double, and so there are rounding errors.
```
filiph: Why remove this?

lib/src/soloud.dart
```diff
@@ -1017,17 +1037,11 @@ interface class SoLoud {
   /// to `1.0` (meaning full volume).
   ///
   /// Throws [SoLoudNotInitializedException] if the engine is not initialized.
-  ///
-  /// Note that if you `setGlobalVolume()` to `0.8` and then
-  /// `getGlobalVolume()`, you might get a slightly different number,
-  /// such as `0.800000042353`.
-  /// This is expected since the internal audio engine uses float
-  /// instead of double, and so there are rounding errors.
```
filiph: Why remove this?

lib/src/soloud_capture.dart
filiph, line 1: I'm only breezing through this file very quickly since the API is experimental.

I know I keep harping on about it, but I think we should take the time and make the very hard decisions of dropping some of this functionality. This is just way too much code to be maintained in the long term, even by a team of volunteers (and right now, it's mostly just you, Marco, to be honest, although I try to help).

lib/src/utils/loader.dart
```diff
@@ -1,303 +1,2 @@
...
-
-  @override
-  bool operator ==(Object other) {
-    return other is _TemporaryFileIdentifier &&
-        other.source == source &&
-        other.path == path;
-  }
-}
+export 'package:flutter_soloud/src/utils/loader_io.dart'
+    if (dart.library.html) 'package:flutter_soloud/src/utils/loader_web.dart';
```
filiph, line 2: [nit] It would be a bit cleaner to have a stub implementation here, like discussed in https://dart.dev/guides/libraries/create-packages#conditionally-importing-and-exporting-library-files.

Something like

```dart
export 'loader_base.dart' // Stub implementation
    if (dart.library.io) 'loader_io.dart' // dart:io implementation
    if (dart.library.js_interop) 'loader_web.dart'; // package:web implementation
```

The stub implementation is an interface class, something like

```dart
interface class Loader {
  bool get blah => throw UnsupportedError('platform not supported');

  void doSomething() => throw UnsupportedError('platform not supported');
}

```

filiph: This is not a big deal, though. Maybe just add a TODO comment with the link.

lib/src/worker/worker.dart
filiph, line 1: Full disclosure: at this point, I probably know much less about `js_interop` than you do.

lib/src/worker/worker.dart
```diff
@@ -0,0 +1,117 @@
+// ignore_for_file: public_member_api_docs
+
+import 'dart:async';
+import 'dart:convert' show jsonEncode;
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+// ignore: avoid_web_libraries_in_flutter
+import 'dart:js_util';
```
filiph, line 8: Why this ignore? Can you maybe add a comment to the file that explains why we need this and how we can potentially, in the future, get rid of this?

lib/src/worker/worker.dart
```diff
@@ -0,0 +1,117 @@
...
+    controller._worker?.onmessage = ((web.MessageEvent event) {
+      controller._outputController?.add(event.data.dartify());
+    }).toJS;
+
+    return controller;
+  }
+
+  /// Set the worker created in WASM.
+  /// This is used to get events sent from the native audio thread.
+  void setWasmWorker(web.Worker wasmWorker) {
```
filiph, line 78: I feel like I need a diagram somewhere in the docs (maybe in `WEB_NOTES.md`?). What are the different workers (WASM worker, this worker), how do they communicate, with whom, etc. Either an ASCII graph or even just a piece of paper that you take a photo of.

web/compile_wasm.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
```
filiph, line 2:
```suggestion
#!/bin/bash
set -euo pipefail

```

Just some bash defensive programming: https://buildkite.com/docs/pipelines/writing-build-scripts#configuring-bash

wasm.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
filiph, line 2:
```suggestion
#!/bin/bash
set -euo pipefail

```

web.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
filiph, line 2:
```suggestion
#!/bin/bash
set -euo pipefail

```

web/compile_worker.bat
```diff
@@ -0,0 +1,2 @@
+ # -O4
+dart compile js -o worker.dart.js ./worker.dart
```
filiph, line 2:
```suggestion
 # -O4
dart compile js -o worker.dart.js ./worker.dart

```

Just adding newline at EOF.

web/compile_worker.sh
```diff
@@ -0,0 +1,2 @@
+ # -O4
+dart compile js -O3 -o worker.dart.js ./worker.dart
```
filiph, line 2:
```suggestion
 # -O4
dart compile js -O3 -o worker.dart.js ./worker.dart

```

web/worker.dart.js.deps
filiph, line 1: This file looks machine-specific (it has paths with `/home/deimos`). Maybe gitignore it?

filiph (approved):
LGTM! Comments are minor.

Let's wait a few days in case there are others who'd like to chip in. But otherwise, let's merge and go from there.

lib/src/bindings/audio_data_web.dart
alnitak: I would like to think that this plugin is not only useful for game developers but also be used for example for a music player. In 2019 when I released my music app, my greatest enjoyment was providing users with an audio visualizer in OpenGL!
IMHO giving developers a way to capture audio data could result in something niche but also something unique. In your game, for example, you could think of making the vertices of the widget at the bottom left move based on system sounds (shots, movements, etc.). 
Not much more than what is already implemented can be added and as in the case of waveforms, we could leave the state of this feature frozen.

filip

lib/src/soloud.dart
```diff
@@ -422,12 +426,25 @@ interface class SoLoud {
   /// The [buffer] represents the bytes of an audio file. It could be also a
   /// simple WAV format sequence of manually generated bytes.
   ///
-  /// It is useful when using this plugin on the web browsers because
-  /// they cannot read directly files in the loal storage.
+  /// When [mode] is [LoadMode.memory], the whole uncompressed RAW PCM
```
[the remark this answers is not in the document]
alnitak: Oh yes you are right. The same parameter is used also in the `loadFile()` method. I just copyed it.
This method acts as the `loadFile()`. It accepts the `buffer` as a compressed audio file bytes and then manage it in a common way. Maybe I have to write this in the doc.

lib/src/soloud_capture.dart
alnitak: 😕 I understand your feeling about this. For game development it is useless! For the same idea of ​​considering this plugin not only for game development, but also possibly for other things, could we consider this as experimental-freezed-locked-`forbidden to ask for improvements` feature? 😀
I know you are definitely right. And in any case, your help has been and will be fundamental and I will never stop thanking you!

pubspec.yaml
alnitak: These assets take 626 Kb uncompressed. It's ok also for me to leave everything here. I just add a comment for those 2 issues to remember this.

web/compile_wasm.sh
alnitak: pretty sweet!

web/worker.dart.js.deps
alnitak: Ouch! I have some strange nicknames don't I? :)

lib/src/soloud.dart
alnitak: I still have that note! Don't know, but on GitHub I see this
aaa
I have that for `[s|g]etGlobalVolume()` and `[s|g]volume()`. Maybe you don't have pulled something?

alnitak:
> > IMPORTANT: loadMem() with mode LoadMode.memory used the on web platform will freeze the UI for the time needed to decompress the audio file. Please use it with mode LoadMode.disk or load your sound when the app starts.
> 
> This is clearly not something we'll address in this PR, but I wonder why this happens. I thought the audio decompression happens in C, and therefore on a non-UI thread, correct?

When you call a C function from Dart it is executed in the main Flutter UI thread, but if that function creates another thread to do stuff, it doesn't affect the UI thread. 
Here the problems are 2 I think. The 1st is that the decoding stuff is made in the main UI thread. The 2nd is that the decoding stuff blocks also the audio Worker (or Worklet)!
In fact on the other platforms, even if you voluntarily block/freeze the UI (ie stop at a breakpoint), the sound still plays.

I need more time to address this and don't know if I'll be able to.

lib/src/worker/worker.dart
alnitak: `jsSendMessage()`, `callbackToStream()` and `class Worker` are only used by `web/worker.dart`. Beeing it stored in the `lib/worker` dir, the linter thinks this source is used for deployment on other platforms too, but it's not.
It's worth to try moving these in the `web/worker.dart` code. I will try before merging.

Anyway, we should also rake a note that `dart:html` will be [deprecated in the future](https://dart.dev/interop/js-interop). We use it in the `Loader` for example. Likewise the `dart:js_util` used here but I didn't find a way to use js_interop instead.

lib/src/worker/worker.dart
alnitak: Don't think so! I still think JS is something to eat :D

lib/src/worker/worker.dart
alnitak: Sure, I'll try!

lib/src/soloud_capture.dart
filiph: Absolutely. Just one thing — if there's another time when you find yourself doing a lot of work on this file just to keep something working, let's revisit this discussion. Of course I'm very happy for every feature that `flutter_soloud` supports, so as long as it's "free" (in terms of our work) or almost free, I'm only happy to have it.

I also like the "experimental-freezed-locked-forbidden to ask for improvements feature" policy. Let's keep that.

lib/src/soloud.dart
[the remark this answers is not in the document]
filiph: Thanks for reminding me!

lib/src/soloud.dart
filiph: Oh, interesting. Maybe some GitHub bug. I was seeing this in the GitHub review UI. I'll have another look but I'll assume it's fine for now.

lib/src/bindings/audio_data_web.dart
filiph: No yeah, I get it. I'm just comparing this with the massive use case of "simply" playing audio. If you're fine with supporting this, of course I'm fine with it, too. Just please tell us long in advance if you ever feel like you start burning out of maintaining this project. It's "only" been 1 year. There are many years to come. :)

lib/src/worker/worker.dart
filiph: Got it.

If anyone with `js_interop` experience is reading this PR, then — first of all, hi! Thanks for having a look! If you can, please do review this file in particular.

alnitak: I moved from here what was needed only in `web/worker.dart`. The `class Worker`, `callbackToStream()` and `jsSendMessage()` are now in `web/worker.dart`.

Also the `dart:js_util` has been replaced by `dart:js_interop_unsafe` and the code adjusted to support it. The ignore is now gone.

lib/src/worker/worker.dart
alnitak: I drew a sketch [here](https://excalidraw.com/#room=09413bf7417b2abd5b22,jJNb4nLgpMKf4vqOXZvmcA). It would be nice to have your opinion before exporting it as an image to insert it in WEB_NOTES. Feel free to modify it!

filiph: This is really helpful! I don't think I have anything to change. It would be great to have things linked (like being able to actually see the EM_ASM() code, or linking to explanatory articles about WASM Modules), but I have no idea how we could do this without making the docs unnecessarily complicated. I think the current version is excellent.

pubspec.yaml
asaarnak: @filiph @alnitak 
This is now released with the new Flutter 3.41.0 😍
https://github.com/flutter/flutter/pull/176393

PR #105, opened 2024-07-15
player initialization params, voice groups, filters for single sounds.

### LINKED ISSUE #78: perf: Ability to play precise sound sequences
**Description**

When I tried to play sharp short sounds very quickly I noticed that the distance between the sounds are not constant, they get messy. A quick way to test this is with the example project, is to replace  'assets/audio/explosion.mp3'  to this [Rimshot1.wav.zip](https://github.com/alnitak/flutter_soloud/files/15036372/Rimshot1.wav.zip) and play them every 100ms. 

```
  Future<void> playTenTimes() async {
    for (int i = 0; i < 20; ++i) {
      await Future.delayed(const Duration(milliseconds: 100));
      await playAnotherInstance();
    }
  }
```
I know this package is not a music sequencer, but it would be awesome, if we could use it for that as well. In the SoLoud documentation I found that `Soloud.playClocked()` addresses something similar, but I am not sure it is the same issue.

### LINKED ISSUE #92: feat: Voice Groups
**Description**

[Voice Groups](https://solhsa.com/soloud/voicegroups.html) is a very useful feature for games. It allows controlling multiple voices in the same time. It would save game developers quite a lot of writing boilerplate, and s the above linked page says it fixes any synchronization issues.

**Requirements**

- [ ] Add support for creating voice groups, adding voices to it, and destroying the voice group. Treat handles and voice group handles as one to control groups properly.

**Additional Context**

It would be greatly appreciated if this is looked into.
P.s: Please let me know if I am creating more issues than necesary, but we are developing a game that heavily relies on audio, and this library of great help. I am creating issues for what we find. But let me know if you want me to create a bigger issue with all the features/problems we face or if an issue for each could help you track them better. Thanks

### LINKED ISSUE #100: use case: metronome
I have yet to find a flutter package that can keep time accurately enough for use as a metronome. I think we should add such an example to flutter_soloud to both ensure the package maintains best-in-industry playback scheduling accuracy, but to also show people what techniques are needed to play with accuracy

### LINKED ISSUE #102: docs: Some questions about gapless looping audio.
**Description**

What's the recommended approach for gapless looping ambient audio? What file format should be used? And is there anything else that needs to be considered? 

Reason I'm asking is, right now, my app uses just_audio with m4a files. The m4a files have been processed to ensure their number of samples is a multiple of 1024, which makes them seamlessly loopable. But m4a file format is not supported by flutter_soloud, mp3 causes gaps when looping, and as far as I'm aware of, iOS doesn't support ogg, making me wonder how to accomplish seamlessly looping sounds.

Any advise is appreciated, 
Thank you.

**Requirements**

- [ ] Requirements go here

alnitak:
## Description

- added player initialization parameters resolves #102, resolves #100, resolves #78:
  - sample rate
  - buffer size
  - number of channels (mono, stereo, quad, 5.1, 7.1)
  
- added `Voice groups` resolves #92:
  - createVoiceGroup()
  - destroyVoiceGroup()
  - addVoicesToGroup()
  - isVoiceGroup()
  - isVoiceGroupEmpty()

- it's now possible to set filters not only globally, but also to single audio sources #101:
  - the `AudioSource` class has now
    - isFilterActive()
    - addFilter()
    - removeFilter()
    - setFilterParameter()
    - getFilterParameter()
    - fadeFilterParameter()
    - oscillateFilterParameter()

- `fadeGlobalFilterParameter()` and `oscillateGlobalFilterParameter()` added also to `SoLoud` for global filters

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/bindings/bindings_player.dart
```diff
@@ -535,7 +591,45 @@ abstract class FlutterSoLoud {
...
+  ///
+  /// [handle] the handle of the voice to apply the fade. If equal to 0,
+  /// it fades the global filter.
+  /// [filterType] filter to modify a param.
+  /// [attributeId] the attribute index to fade.
+  /// [to] value the attribute should go in [time] duration.
+  /// [time] the fade slope duration.
+  /// Returns [PlayerErrors.noError] if no errors.
+  @mustBeOverridden
+  PlayerErrors fadeFilterParameter(
```
alnitak, line 606: These methods now accept also the `SoundHandle` or the `SoundHash`. These parameters are needed by the callers from `AudioSource` when managing filters. When called from `class SoLoud` these parameters are not needed because the filters are managed globally (for all the sounds).
In short, `AudioSource` manages filters for single sounds/voices and `SoLoud class` global filters.

filiph:
Should this be merging into `web`?

alnitak:
> Should this be merging into `web`?

Ouch, I made a mistake I think. This should be merged to `main`. Is there a way to fix?

sbauly:
Hi @alnitak, this PR looks very exciting! Regarding support for non-global filters, will we also be able to apply filters to voice groups? I imagine that may be useful in the case of wanting to apply a filter to all sounds apart from one or two. For example, a game might have a Freeverb filter for all in-game sounds, but not for the menu sfx.

Also, I'm not sure it's possible, but would the filters be able to behave similarly to buses/sends? This would probably be the most ideal way of using filters, as very few instances would be needed whilst still offering flexibility for different use cases. It would be useful to have a Freeverb send that SoundHandles could be sent to in varying amounts of wetness, for example.

alnitak:
Hi @sbauly, thanks for the interest.

> will we also be able to apply filters to voice groups? 

I don't think it is somehow feasible. In the groups laid different voice handles that may belong to different `AudioSource`s.
A filter (or more than one) can be added only to an `AudioSource` (not to voice handles it owns) and all its handles are affected. Although you can set different filter parameters for each handle of an `AudioSource`.
So I suppose this logic should be made outside the plugin.

> Also, I'm not sure it's possible, but would the filters be able to behave similarly to buses/sends? This would probably be the most ideal way of using filters, as very few instances would be needed whilst still offering flexibility for different use cases. It would be useful to have a Freeverb send that SoundHandles could be sent to in varying amounts of wetness, for example.

Don't know if I understand correctly, but maybe I answered above to this. Anyway, this is the logic:

```
AudioSource source = loadFile(...);
source.addFilter(Freeverb);
SoundHandle handle1 = play(source);
SoundHandle handle2 = play(source);

/// now you are able to set parameters to handles
source.setFilterParameter(handle1, Freeverb.wet, 0);
source.setFilterParameter(handle2, Freeverb.wet, 1);
```

I hope I have been clear and helpful

lib/src/audio_source.dart
```diff
@@ -103,6 +122,186 @@ class AudioSource {
...
+    return error;
+  }
+
+  /// Set the effect parameter with id [attributeId] of [filterType]
+  /// with [value] value.
+  ///
+  /// Specify the [attributeId] of the parameter (which you can learn from
+  /// [SoLoud.getFilterParamNames]), and its new [value].
+  ///
+  /// [handle] the handle to set the filter to. If equal to 0, the filter is
```
filiph, line 182: This is weird. Either allow `null`, and then translate the null handle to `0` internally, or create two different methods (one for global filter and another one for single-voice filter).

lib/src/bindings/bindings_player_web.dart
```diff
@@ -420,79 +428,182 @@ class FlutterSoLoudWeb extends FlutterSoLoud {
...
+  /////////////////////////////////////////
+  /// voice groups
+  /////////////////////////////////////////
+
+  @override
+  SoundHandle createVoiceGroup() {
+    /// The group handle returned has the sign bit flagged. Since on the web
+    /// the int is a signed 32 bit, a negative number will be returned.
+    /// Fixing by ORing the result.<
+    final ret = wasmCreateVoiceGroup() | 0xfffff000;
```
filiph, line 440: This is a bit scary: can this become a source of bugs when some internal mechanism changes? I'd prefer to at least log the original return value and the fact that we're changing it. So that if we ever get a bug from this, we at least have a way to find the problem in the logs.

lib/src/soloud.dart
```diff
@@ -1333,6 +1350,57 @@ interface class SoLoud {
     _controller.soLoudFFI.setFftSmoothing(smooth);
   }
 
+  /////////////////////////////////////////
+  /// voice groups
+  /////////////////////////////////////////
+
+  /// Used to create a new voice group. Returns 0 if not successful.
```
filiph, line 1357: In Dart, failure that likely has effect on the following code (like here) should generally throw. Alternatively, return `null`. In both cases, the developer is forced to deal with the possibility of failure (fail fast) instead of mistakenly trudging on.

lib/src/soloud.dart
```diff
@@ -1333,6 +1350,57 @@ interface class SoLoud {
...
+  /// voice group is empty.
+  ///
+  /// [handle] the group handle to check.
+  /// Return true if [handle] is a group handle.
+  bool isVoiceGroup(SoundHandle handle) {
+    return _controller.soLoudFFI.isVoiceGroup(handle);
+  }
+
+  /// Checks whether a voice group is empty. SoLoud automatically trims
+  /// the voice groups of voices that have ended, so the group may be
```
filiph, line 1395: Just to be clear: does the trimming happen also in SoLoud C++? I'd prefer we stay as close to the underlying library as possible in this respect.

example/tests/tests.dart
```diff
@@ -676,3 +645,186 @@ Future<StringBuffer> testSynchronousDeinit() async {
...
+
+/// Test voice groups.
+Future<StringBuffer> testVoiceGroups() async {
+  await initialize();
+
+  final currentSound =
+      await SoLoud.instance.loadAsset('assets/audio/explosion.mp3');
+
+  /// Start playing sounds in pause state to get their handles.
+  final h1 = await SoLoud.instance.play(currentSound, paused: true);
```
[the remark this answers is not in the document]
alnitak: I thought the same at the beginning, but groups accepts only the handles returned by the `play`. When the handle ends it is also removed from the group. It's not possible to automatically add handles of a `SoundHash` using the C++ SoLoud lib, but maybe something can be done outside it (ie a new feature to use sounds instead of handles in groups).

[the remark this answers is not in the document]
filiph: Nono, this is fine, then. Let's keep close to the original API in this case.

lib/src/audio_source.dart
alnitak: In `SoundHandle` fortunately we accept a value of 0, but this 0 is not handled by SoLoud c++ (only > 0). I used this only on the FFI side and this comment is wrong and you catched me! :) (I copied/pasted wrongly I think). It should be just:
`/// [handle] the handle to set the filter to.`

Here we are in the `AudioSource` and that `handle` is mandatory. While in `SoLoud.setGlobalFilterParameter()` is not a parameter but they both call the same FFI method.

lib/src/bindings/bindings_player_web.dart
alnitak: A negative number is always returned on Web and drove me crazy for many half hours! :)
Now, I don't remember well what were under the hood of JS and C++/WASM, but I tested a lot and anyway I think we will never touch the internal SoLoud C++ groups.
After saying this, there are no problems for me to log the original `wasmCreateVoiceGroup()` return value if you want.

lib/src/soloud.dart
alnitak: Yes, I took this doc from the SoLoud C++ html page.

lib/src/soloud.dart
alnitak: Right! I'll throw a new exception.

src/bindings.cpp
```diff
@@ -977,24 +1036,34 @@ extern "C"
...
-        *index = player.get()->mFilters.isFilterActive(filterType);
+
+        if (soundHash == 0)
+            *index = player.get()->mFilters.isFilterActive(filterType);
+        else
+        {
+            auto const s = player.get()->findByHash(soundHash);
+            if (s == 0)
+                return soundHashNotFound;
+            *index = s->filters->isFilterActive(filterType);
```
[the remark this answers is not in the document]
alnitak: Oh yes, nullptr is under the hood =0. For clarity and since we are no more in the '80s, it would be better to use nullptr.

PR #110, opened 2024-07-29
chore

### LINKED ISSUE #98: feat: set relative speed without changing the pitch
**Description**

I would like to change the relative speed of the playing sound without affecting the pitch, similar to [tempo change in Audacity](https://manual.audacityteam.org/man/change_speed.html) or YouTube player speed.

**Requirements**

- [ ] A way to change audio speed/tempo without changing the pitch of the audio

**Additional Context**

Currently the `setRelativePlaySpeed` changes the pitch if you change the speed

### LINKED ISSUE #108: feat: make filters easier to use
**Description**

As for now methods like `setFilterParameter()` have parameters like `int attributeId` which identifies the index number to the filter parameter. This is difficult to choose since parameters are stored in records for each filter. For example, Freeverb parameters are stored in `fxFreeverb` in `filter_params.dart` and are accessible only by knowing the index and reading the code.

Would be nice to have classes easily accessible to get parameter names, mins, maxs, etc for each filter.

alnitak:
## Description

---
Removed the main example which was too much complicate for a first approach. In the futere I'll make a new GitHub repository with that. 
For now there are 2 little example:
- a metronome
- a filter use case

---
Added `pitch shift` filter to somehow resolve #98.
This filter has these parameters:
- `shift`: a value of 0.5 means one octave down (12 semitones down), a value  of 2.0 means one octave up (12 semitones up).
- `semitones`: increases or decrease using semitones (¹/12 of an octave).

This filter can be used in conjunction with `SoLoud.setRelativePlaySpeed()` to for example play a sound with voices at double speed without changing the tone speech.

---
The filters can now be used for single sounds instead of using them only globally.

**IMPORTANT**: filters for individual sounds are not supported on the
web platform. On the other platforms the filter must be added before playing. Only voice handles played after adding the filter will play with the filter chosen.

Filters can now be accessed globally using `SoLoud.instance.filters` or for single [AudioSource] sounds with `sound.filters` to resolve #108. With something like:
```
/// For global filters.
final ps = SoLoud.instance.filters.pitchShiftFilter;
ps.activate();
ps.shift(soundHandle: soundHandle).value = 0.6; // set value.
final shift = ps.shift(soundHandle: soundHandle).value; // get value.
ps.fadeFilterParameter(to: 3, time: const Duration(milliseconds: 2500));
ps.oscillateFilterParameter(from: 0.5 to: 1.5, time: const Duration(milliseconds: 2500));
ps.queryShift.[min | max | def] // to get filter min, max and default values.

/// For single sound filters.
final ps = sound.filters.pitchShiftFilter;
[as for global]

```

Look at the respective documentation to see how to use them.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


alnitak:
For this PR we need to choose the plugin version, the check fails on the CHANGELOG.md which doesn't have it.  On pub.dev this plugin has not been updated to the main branch since PR [85](https://github.com/alnitak/flutter_soloud/pulls?q=is%3Apr+is%3Aclosed) so we may think of releasing 2.1.0 or a major 3.0 since the news are important.

For this review, regarding the new filters' way to manage them, in the `filters` dir there are 10 source files for the same number of filters. These sources differ only in their different parameter names, so looking only at one will be enough I think.

`filters/filters.dart` is the source where those filters are managed. There is a `FiltersSingle` class which is instantiated as `AudioSource.filters` class and a `FilterGlobal` class which is instantiated as `SoLoud.filters`.

filiph:
LGTM! Thanks!

filiph:
Also verified the examples on macOS. Love the new `tests/tests.dart` test app, btw!

PR #116, opened 2024-08-20
Use C++ 17 to build on iOS

kirusfg:
## Description

Forcing the use of C++ 17 when building for iOS, as 46c6f4bd2d816eff58ca4033cb4b6c4ce9886bf1 does for Android and Windows. The XCode toolchain on MacOS seems to use C++ 11 by default for compilation and the plugin fails to be built. It was confirmed that the addition of this flag helps in building the plugin successfully.  

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/ios/flutter_soloud.podspec
+++ b/ios/flutter_soloud.podspec
@@ -27,7 +27,8 @@ Flutter audio plugin using SoLoud library and FFI
     '-w',
     '-DOS_OBJECT_USE_OBJC=0', '-Wno-format',
     '-lpthread',
-    '-lm'
+    '-lm',
+    '-std=c++17'
   ]
   # Flutter.framework does not contain a i386 slice.
   s.pod_target_xcconfig = { 


alnitak:
Thanks @kirusfg,

this was a fix I've made in the `pitch` branch ([here](https://github.com/alnitak/flutter_soloud/pull/110/commits/0a53b127dd71ecbef800135e6635d1f1fdbbb1b5#diff-0d9d6fe1dae3e7e03a42e34645fb956144aa99ad36d56e281337d58aecb2af68)), but I did in a little different way:
```
  s.pod_target_xcconfig = { 
    ...
    "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
    "CLANG_CXX_LIBRARY" => "libc++"
  }
```
Since I am not familiar with Mac and iOS, do you know if that is the right way? And also, your change could coexist with the above I did?
If yes, could you please join what you did with my fix?

Another little thing. I think this should be done also for the MacOS podspec. Can you please do the same before I merge?

Thanks a lot for your contribution!

kirusfg:
I'm not familiar with the Apple ecosystem myself - me and my boss who has a Mac have simply discovered that the latest branch does not build for iOS yesterday. Setting the C++ standard was the fix.

I don't think there would be anything specific to Mac when it comes to defining the C++ standard. I think as long as both Clang and GCC are instructed to use `c++17`, it would work.

Me and my colleague will see if your fix works on iOS and MacOS today.

alnitak:
Version 2.1.0 has landed with the fix in the `pitch` branch.

Anyway, thank you so much @kirusfg!

Issue #117, opened 2024-08-21
fix: Possible crash with EXC_BAD_ACCESS KERN_INVALID_ADDRESS at 0x1166800a0


filiph:
**Description**

I got the following exception, reported by Sentry. I can't vouch for the bug since it was raised in debug (so maybe it's something that happened because of hot reload or something like that). But I file it anyway in case you have ideas.

![Screenshot 2024-08-21 at 20 33 21](https://github.com/user-attachments/assets/dc293f1d-3ca2-4a4d-bb66-f6d3ca0cad6f)

![Screenshot 2024-08-21 at 20 34 02](https://github.com/user-attachments/assets/9a2f3403-409a-4e93-a502-151499c75c4d)

Actually, now that I'm looking at the "breadcrumbs" (basically log history), it seems that this might have happened when I closed the app and the app tried to play a sound (the click sound of the "EXIT" button) while the app was being shut down.

Not sure if this is addressable, so feel free to close this

**Steps To Reproduce**

1. Try to create a condition where the app is `exit()`ing while a `SoLoud::on_notification` is called.

**Expected Behavior**

Event is ignored because SoLoud is (hopefully?) being shut down or is already shut down.

alnitak:
Oh WOW! What makes me think is the first line of your screenshot: "Crashed in non-app", which is odd I think.

My thought is that a hot reload does this:
1. the `SoLoud.init()` method is called just after a hot reload
2. since the `init()` method sees the player is already initialized, it will `deinit()` it
3. meanwhile the state changes and the `SoLoud::on_notification` didn't realize yet a new address of the newly initialized player

Or maybe after a hot reload the Flutter main isolated is terminated and re-executed, but the audio thread is not yet. So the state changes and the old audio thread fires the callback that doesn't exist anymore.

I would leave this issue open for a while to see also if that state notification can be useful in the future because now it's not.

Thanks for your finding.

alnitak:
I tried to use Sentry and I didn't see what you got. At least on Linux and Android.

Thinking about this, I created the branch `state_changed` that hopefully could fix this.

@filiph can you please try to use that branch to see if the error persists?

filiph:
The thing is, I don't think I can reliably reproduce the issue. I didn't even notice it until seeing it much later in Sentry. I'm pretty sure the app closed just fine from the perspective of the user.

Feel free to close this for now. At the very least, it's good to have it here for SEO so that people searching for "EXC_BAD_ACCESS soloud" will get here.

PR #119, opened 2024-08-26
added query filter status

### LINKED ISSUE #118: Check filters status
@alnitak, How do you now track filter statuses between hot restarts?
i.e. I'm following codelab from google io 2024, and there were a line that we could check status with `isFilterActive(filter)` method, which is now deprecated.

_Originally posted by @misterfourtytwo in https://github.com/alnitak/flutter_soloud/issues/108#issuecomment-2309004690_

alnitak:
## Description

- Addded query filter status to fix #118 ie:
```
bool isEchoActive = SoLoud.instance.filters.echoFilter.isActive();
```

- Chore: reset callback function pointers on dispose on C side. Probably unnecessary and irrelevant but it could solve #117

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


filiph:
I have a single comment (about naming of `isActive()`) and two nits.

alnitak:
> I have a single comment (about naming of `isActive()`) and two nits.

Thanks to your valuable suggestion, I added the abstract class `FilterBase` common to all filter classes.
Also, the method docs are now visible!

I also forgot to add a `timeStretch()` method in the `PitchShiftSingle` filter. It just calculates the value for the `shift` based on the requested speed (it acts like 1.5x, 2x etc from YouTube without changing the timbre of the voice).

lib/src/filters/bassboost_filter.dart
```diff
@@ -23,19 +23,12 @@ enum BassBoostEnum {
       };
 }
 
-abstract class BassBoostInternal {
-  const BassBoostInternal(SoundHash? soundHash) : _soundHash = soundHash;
+abstract class BassBoostInternal extends FilterBase {
```
filiph, line 26: [nit] If these `SomeFilterInternal` classes really are supposed to be internal, then I'd suggest:

1. Making the classes private (by renaming to `_BassBoostInternal`), OR
2. If they cannot be made private (i.e. only accessible within the same file), then maybe mark them as `@internal` or `@visibleForTesting`

If they're more like Base classes (publicly visible), then they could be renamed to something like `BassBoostBase`.

filiph (approved):
This is great! LGTM.

I added a suggestion but if you want, you can do that later or never.

lib/src/filters/bassboost_filter.dart
alnitak: done.

PR #131, opened 2024-09-18
Fix waveform and iOS compilation issue

### LINKED ISSUE #128: iOS build fails with: Semantic Issue (Xcode): 'path' is unavailable: introduced in iOS 13.0
First things first: Thank you very much for providing flutter_soloud.

**Description**

Trying to run an app with flutter_soloud version 2.1.3 on iOS (17.5.1) fails with:
```
Launching lib/main.dart on Somebody’s iPad in debug mode...
Automatically signing iOS for device deployment using specified development team in Xcode project: XXXXXXXXXX
Xcode build done.                                           19.3s
Failed to build iOS app
Could not build the precompiled application for the device.
Semantic Issue (Xcode): 'path' is unavailable: introduced in iOS 13.0
/Users/.../ios/.symlinks/plugins/flutter_soloud/src/bindings.cpp:250:25

Semantic Issue (Xcode): 'u8path<char *>' is unavailable: introduced in iOS 13.0
/Users/.../ios/.symlinks/plugins/flutter_soloud/src/bindings.cpp:250:52

Semantic Issue (Xcode): '~path' is unavailable: introduced in iOS 13.0
/Users/.../ios/.symlinks/plugins/flutter_soloud/src/bindings.cpp:250:35

Semantic Issue (Xcode): '~path' is unavailable: introduced in iOS 13.0
/Users/.../ios/.symlinks/plugins/flutter_soloud/src/bindings.cpp:250:30

Semantic Issue (Xcode): 'string' is unavailable: introduced in iOS 13.0
/Users/.../ios/.symlinks/plugins/flutter_soloud/src/bindings.cpp:254:44
2

Error launching application on Sombody’s iPad.

Exited (1).
```

**Additional Context**

Looks like this got introduced in version 2.1.0. Version 2.0.2 builds just fine on iOS.

### LINKED ISSUE #129: fix: Issue with Frequency Change in Waveform: When Changing Frequencies, Background Frequency Persists and High Frequencies Take Time to Generate
Description

When changing the frequency of waveform sounds using the flutter_soloud package, after a while, some background frequencies are heard along with the original frequency. Additionally, when changing to high frequencies (above 10,000 Hz), the engine struggles and takes extra time to generate the sound.

Steps To Reproduce

1. Use `SoLoud.instance.loadWaveform(WaveForm.sin, true, 1, 0)` to generate a sine wave sound.
2. Change the frequency of the generated sound using: `SoLoud.instance.setWaveformFreq(soundHandle, frequency);`
3. Notice that after the frequency change, a lower frequency can be heard in the background with the original frequency.
4. High frequencies (above 10,000 Hz) take longer to generate, and there’s a delay in producing sound.

Expected Behavior

The frequency change should happen without any background or low-frequency sound. High frequencies (above 10,000 Hz) should play immediately without delay.

Please help resolve this issue.

### LINKED ISSUE #130: fix: Issue with "unsupported option '-G' for target 'x86_64-apple-ios13.0-simulator'"
**Description**

**Steps To Reproduce**

1. Update xCode to 16.0 and macOS to 15
2. Use flutter_soloud with version 2.1.3
3. Try to compile iOS project from xcode or vs code on simulator (In my case iPhone 16 Pro)
4. Get error:
```
Error (Xcode): unsupported option '-G' for target 'x86_64-apple-ios13.0-simulator'

Error (Xcode): unsupported option '-G' for target 'arm64-apple-ios13.0-simulator'
```

**Expected Behavior**

App should compile without errors

**Screenshots**
Vscode:
Screenshot 2024-09-18 at 13 30 15

Xcode
Screenshot 2024-09-18 at 13 30 53

**Additional context**
When removing library from project, everything compiles fine

alnitak:
## Description

fix #129 
fix #128
fix #130

## Type of Change

> When changing the frequency of waveform sounds using the flutter_soloud package, after a while, some background frequencies are heard along with the original frequency.

The waveform generation (on C+ side) used float numbers, but higher precision was needed to compute the waveform. Changing to `double`s fixed the issue!

Minor changes for the other 2 issues that fix compiling on iOS.

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
             "name": "Flutter debug",
             "type": "dart",
             "request": "launch",
-            "program": "lib/filters/pitchshift.dart",
+            "program": "lib/main.dart",
             "flutterMode": "debug",
             "cwd": "${workspaceFolder}/example"
         },
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -9,7 +9,7 @@
         },
         {
             "label": "compile linux debug",
-            "command": "cd ${workspaceFolder}/example; flutter build linux -t lib/filters/pitchshift.dart --debug",
+            "command": "cd ${workspaceFolder}/example; flutter build linux -t lib/main.dart --debug",
             "type": "shell"
         },
         {
@@ -29,7 +29,7 @@
         },
         {
             "label": "compile web debug",
-            "command": "cd ${workspaceFolder}/example; flutter run -d chrome --web-renderer canvaskit --web-browser-flag '--disable-web-security' -t lib/audio_data/audio_data.dart --release",
+            "command": "cd ${workspaceFolder}/example; flutter run -d chrome --web-renderer canvaskit --web-browser-flag '--disable-web-security' -t lib/main.dart --release",
             "type": "shell"
         }
     ]
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -39,8 +39,6 @@ PODS:
   - path_provider_foundation (0.0.1):
     - Flutter
     - FlutterMacOS
-  - permission_handler_apple (9.1.1):
-    - Flutter
   - SDWebImage (5.16.0):
     - SDWebImage/Core (= 5.16.0)
   - SDWebImage/Core (5.16.0)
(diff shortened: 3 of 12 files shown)


filiph:
[nit] I think raising deployment target is definitely a breaking change. In the future, we should consider only doing it between major semver updates. This change broke the codelab and therefore might have broken others' code, too.

(Just to be clear, I'm glad these fixes landed regardless, and having the iOS version at `13` is probably good future-proof-wise. Just a heads up for the next time we need to change deployment targets.)

alnitak:
Sorry, I had to think of that! 🫣

PR #141, opened 2024-10-22
Choose output playback device

### LINKED ISSUE #142: feat: Choose output playback device
**Description**

Implement a feature that lets the user choose and change the output playback device.

To get all output devices, use `listPlaybackDevices` which returns a list of `PlaybackDevices` class. Each item of this class contains the id, whether it's the default device (the one used by the OS) and the name.
This method can be called even if the engine has not been initialized.

At any time it is possible to pass to `changeDevice` a `PlaybackDevice` which will change the output device.

The `SoLoud.init` method, now optionally accepts also the `device` parameter.

alnitak:
## Description

To get all output devices use `listPlaybackDevices` which returns a list of `PlaybackDevice`s class. Each item of this class contains the id, whether it's the default device (the one used by the OS) and the name.
This method can be called even if the engine has not been initialized.

At any time it is possible to pass to `changeDevice` a `PlaybackDevice`
which will change the output device.

The `SoLoud.init` method, now optionally accepts also the `device` parameter. If not provided the OS default is used.

Example in `example/output_device/output_device.dart`.

close #142 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


example/lib/output_device/output_device.dart
```diff
@@ -0,0 +1,127 @@
...
+            /// When changing the output device, we need to reinitialize
+            /// the player. All existing audio sources will be stopped and
+            /// disposed as well.
+            // currentDevice = devices[value!];
+            // SoLoud.instance.deinit();
+            // await SoLoud.instance.init(deviceId: currentDevice.id);
+
+            // currentSound = await SoLoud.instance
+            //     .loadAsset('assets/audio/8_bit_mentality.mp3');
+            // await SoLoud.instance.play(currentSound!, looping: true);
```
filiph, line 113: Is this still true? Why is it commented out?

lib/src/enums.dart
```diff
@@ -265,3 +265,23 @@ enum Channels {
   /// The channels count.
   final int count;
 }
+
+/// CaptureDevice exposed to Dart
+final class PlaybackDevice {
```
filiph, line 270: [nit] Not sure why this is in `enums.dart`. Maybe it deserves its own file?

lib/src/soloud.dart
```diff
@@ -358,6 +364,32 @@ interface class SoLoud {
...
   }
 
+  /// Changes the output audio device to the one specified in the [newDevice].
+  /// If [newDevice] is not provided, the default OS device will be used.
+  ///
+  /// Note: Android and Web, only support one output device which is the
+  /// default device (iOS and MacOS?).
+  ///
+  /// Throws [SoLoudNotInitializedException] if the engine is not initialized.
+  PlayerErrors changeDevice({PlaybackDevice? newDevice}) {
```
filiph, line 374: If I remember correctly, I think we stopped returning `PlayerErrors` and instead just throw Dart exceptions. In this case, I think it would make more sense to throw, and then just return `void` (if the change is instantaneous) or `Future<void>` (if it's async).

src/bindings.cpp
```diff
@@ -193,6 +194,65 @@ extern "C"
...
+        {
+            bool hasSpecialChar = false;
+            /// check if the device name has some strange chars (happens on Linux)
+            for (int n = 0; n < 5; n++)
+            {
+                if (d[i].name[n] < 0x20)
+                    hasSpecialChar = true;
+            }
+            if (strlen(d[i].name) <= 5 || hasSpecialChar)
+                continue;
```
filiph, line 228: Please add a comment that explains why we're skipping this device based on this condition.

src/player.cpp
```diff
@@ -33,22 +34,101 @@ void Player::setStateChangedCallback(void (*stateChangedCallback)(unsigned int))
...
     if (mInited)
         return playerAlreadyInited;
 
+    void *playbackInfos_id = nullptr;
+    if (deviceID != -1)
+    {
+        // Calling this will init [pPlaybackInfos]
+        auto const devices = listPlaybackDevices();
+        if (devices.size() == 0 || deviceID >= devices.size())
+            return backendNotInited;
```
filiph, line 48: Shouldn't this be a better error? Something like `noPlaybackDevicesFound` for the empty list, and `playbackDeviceIdOutOfRange` for the other condition?

src/player.cpp
```diff
@@ -33,22 +34,101 @@ void Player::setStateChangedCallback(void (*stateChangedCallback)(unsigned int))
...
+{
+    if (!mInited)
+        return backendNotInited;
+
+    void *playbackInfos_id = nullptr;
+
+    // Calling this will init [pPlaybackInfos]
+    auto const devices = listPlaybackDevices();
+    if (devices.size() == 0 || deviceID >= devices.size())
+        return backendNotInited;
```
filiph, line 76: Same here with the error. One should be `noPlaybackDevicesFound` and the other should be `playbackDeviceIdOutOfRange`.

src/player.cpp
```diff
@@ -33,22 +34,101 @@ void Player::setStateChangedCallback(void (*stateChangedCallback)(unsigned int))
...
+    playbackInfos_id = &pPlaybackInfos[deviceID].id;
+
+    SoLoud::result result = soloud.miniaudio_changeDevice(playbackInfos_id);
+    
+    if (result != SoLoud::SO_NO_ERROR)
+        result = backendNotInited;
```
filiph, line 82: A more descriptive error here, too. Something like `playbackDeviceChangeError`? Or whatever the SoLoud::result can be in this scenario.

src/player.cpp
```diff
@@ -33,22 +34,101 @@ void Player::setStateChangedCallback(void (*stateChangedCallback)(unsigned int))
...
+    // printf("***************** LIST DEVICES START\n");
+    ma_context context;
+    ma_uint32 playbackCount;
+    ma_device_info *pCaptureInfos;
+    ma_uint32 captureCount;
+    std::vector<PlaybackDevice> ret;
+    ma_result result;
+    if ((result = ma_context_init(NULL, 0, NULL, &context)) != MA_SUCCESS)
+    {
+        printf("Failed to initialize context %d\n", result);
```
filiph, line 98: I'm not sure I'm a big fan of just printing to console and returning an empty vector but I guess this deep down it's just a nit.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
```diff
@@ -144,5 +149,27 @@ namespace SoLoud
         aSoloud->mBackendString = "MiniAudio";
         return 0;
     }
+
+    result miniaudio_changeDevice_impl(void *pPlaybackInfos_id)
```
filiph, line 153: Are we sure we want to add these things straight into `soloud_miniaudio.cpp`? Won't this be overridden when we upgrade SoLoud (leading to possibly gnarly merge conflicts)?

src/soloud/src/core/soloud.cpp
```diff
@@ -205,7 +205,24 @@ namespace SoLoud
 		mAudioThreadMutex = NULL;
 	}
 
-	result Soloud::init(unsigned int aFlags, unsigned int aBackend, unsigned int aSamplerate, unsigned int aBufferSize, unsigned int aChannels)
+	// Change output device.
+	// Added by Marco Bavagnoli
+	result Soloud::miniaudio_changeDevice(void *pPlaybackInfos_id)
+	{
+#if defined(WITH_MINIAUDIO)
+		// lockAudioMutex_internal();
```
filiph, line 213: Should these mutex locks be commented out?

filiph (changes requested):
Thanks, this is a really good feature I didn't even know we needed (but we _do_ need it). I have some suggestions.

example/lib/output_device/output_device.dart
alnitak: When I was coding this, I didn't yet know that the device can be changed also after initializing it (forgot to delete!).

lib/src/soloud.dart
alnitak: Fixed and added `SoLoudNoPlaybackDevicesFoundCppException`.

src/bindings.cpp
alnitak: A few lines above I wrote a comment and added few lines now. It happened that some results had the name composed of non-text ASCII characters. Characters with values ​​<0x20 (blank space). For some time now it no longer happens on my Linux, it was probably some issue in the OS, but I preferred to leave this check anyway.

src/player.cpp
alnitak: added CPP `noPlaybackDevicesFound` and counterpart Dart `SoLoudNoPlaybackDevicesFoundCppException`.

src/player.cpp
alnitak: done

src/player.cpp
alnitak: I looked at the code and seems the only error could be `UNKNOWN_ERROR`.

src/player.cpp
alnitak: Replaced with just a comment. If an empty array is returned I suppose is because the hardware doesn't have a playback device.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
alnitak: I tried to see if there was another place. This is the only place where, those who choose to use miniaudio have access to their APIs.

src/soloud/src/core/soloud.cpp
alnitak: removed

src/player.cpp
filiph: What about the other error, where the user is asking for a device ID that doesn't exist? (`deviceID >= devices.size()`)? Shouldn't that have a separate error?

src/player.cpp
filiph: Fair enough, thanks for checking!

filiph: I'd still prefer a more meaningful error or at least a comment here in the code ("miniaudio_changeDevice can only throw UNKNOWN_ERROR"). But that's a nit.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
filiph: Gotcha, thanks for checking!

src/bindings.cpp
alnitak: I'm pretty sure those devices are junk data from `/dev/snd` I got a while back. I don't know how miniaudio gets those devices but in my opinion, they are not valid.

Added some words in the comment.

src/player.cpp
alnitak: I was thinking that either there are no devices or you are asking to use one that does not exist (maybe because the BT headphones have been disconnected), the device you would like to use does not exist, and therefore `noPlaybackDevicesFound` would be enough?
Do you think `playbackDeviceIdOutOfRange` should be there too?

src/player.cpp
alnitak: added a comment

filiph (approved):
LGTM!

hoalarious:
Confirming this doesn't work for android and iOS?

PR #148, opened 2024-11-09
Audio buffer stream

### LINKED ISSUE #144: feat: use PCM audio data to build a sound
**Description**

Would be nice to be able to send PCM data to the player and have the player be able to render it. This could be done by extending the SoLoud C++ [AudioSource](https://solhsa.com/soloud/newsoundsources.html) class to have a `BufferStream` class that can be used as a normal AudioSource sound.

The advantage of this is that the player can be used as a normal [AudioSource] and can play audio data that could be streamed from the web or created directly from Flutter. This new kind of [AudioSource] can then be able to use all the features of the flutter_soloud like filters, faders, oscillators and so on.
Think for example of some AI web services that let you compose voices like [this](https://elevenlabs.io/docs/api-reference/websockets).

**Additional Context**

Would also be nice to have a way to send PCM data directly to the output device without creating a new [AudioSource]: compose some audio PCM in Flutter and send to the player output. But this is maybe for another feature.

alnitak:
## Description

close #144

This new type of `AudioSource` is created by subclassing the SoLoud C++ [AudioSource](https://solhsa.com/soloud/newsoundsources.html) class.

It works by initializing it with its parameters, then adding PCM audio data, and finally signaling that no more data is incoming:
- `setBufferStream` returns the new `AudioSource`. Using its parameters, you can set the PCM audio format, the maximum memory usage, the buffering time required to resume playback (when paused to wait for more data), and a callback to handle buffering events.
- `addAudioDataStream*` adds PCM audio data to the stream. Helper methods are available to add signed/unsigned 8-bit integers, signed 16- and 32-bit integers, and 32-bit floats.
- `setDataIsEnded` marks the end of the data stream.

The returned `AudioSource` is compatible with filters, oscillators and all other tools.

### Use-case
- streaming PCM audio data from a websocket and playing it while receiving data: `example/lib/buffer_stream/websocket.dart` is used for testing. It receives PCM data from a local websocket, provided by the [websocketd](https://github.com/alnitak/websocketd) tool. The Flutter app uses ffmpeg to stream PCM data over the websocket via [websocketd](https://github.com/joewalnes/websocketd), allowing you to simulate slow download speeds and buffering behavior.
- `example/lib/generate.dart` generates audio and fills it into the `AudioBuffer`.
- the [Cartesia TTS Flutter project](https://github.com/alnitak/cartesia_tts) demonstrates how to use Cartesia AI APIs to send text and receive voice audio data.
- you can also use [flutter_recorder](https://github.com/alnitak/flutter_recorder) to capture PCM audio from the microphone and load it into the `AudioSource` for flutter_soloud.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


lib/src/soloud.dart
```diff
@@ -625,6 +658,166 @@ interface class SoLoud {
...
+  ///
+  /// [sampleRate] the sample rate. Usually is 22050 or 44100 (CD quality).
+  ///
+  /// [channels] enum to choose the number of channels.
+  ///
+  /// [pcmFormat] enum to choose from `f32le`, `s8`, `s16le` and `s32le`.
+  ///
+  /// [onBuffering] a callback that is called when starting to buffer
+  /// (isBuffering = true) and when the buffering is done (isBuffering = false).
+  /// It gives back the `handle` which triggered the event and the `time`
```
filiph, line 683:
```suggestion
  /// The callback is called with the `handle` which triggered the event and the `time`
```

lib/src/soloud.dart
```diff
@@ -625,6 +658,166 @@ interface class SoLoud {
...
+  /// in seconds.
+  ///
+  /// Throws [SoLoudNotInitializedException] if the engine is not initialized.
```
filiph, line 686: No other error here? Just checking.

lib/src/soloud.dart
```diff
@@ -625,6 +658,166 @@ interface class SoLoud {
...
+
+    final newSound = _addNewSound(ret.error, '', ret.soundHash.hash);
+    return newSound;
+  }
+
+  /// Add PCM audio data to the stream.
+  ///
+  /// This method can be called within an `Isolate` making possible
+  /// to create PCM data and send them to the buffer without frezing
+  /// the main thread.
```
filiph, line 716: Interesting. Can you add a short in-line example, maybe? Or something like `See the sample code in example/blah.dart`? I don't understand how one can call a method on SoLoud from a different isolate.

lib/src/soloud.dart
```diff
@@ -625,6 +658,166 @@ interface class SoLoud {
...
+    if (e != PlayerErrors.noError) {
+      _logPlayerError(e, from: 'addAudioDataStream() result');
+      throw SoLoudCppException.fromPlayerError(e);
+    }
+  }
+
+  /// Helper method to add `Float32List` PCM audio data to the stream.
+  ///
+  /// This method is a wrapper around [addAudioDataStreamU8].
+  void addAudioDataStramF32(AudioSource source, Float32List audioChunk) {
```
filiph, line 753: I wonder if it's worth it to have all these help methods. It pollutes the API with a bunch of members. Maybe we should just teach people how to "cast" their typed data into raw `UInt8List`? And then you only have one method, `addAudioDataStream()` (without the `U8`).

filiph: And by "teach people how to...", I mean adding sample code into `addAudioDataStream()`'s API doc.

lib/src/soloud.dart
```diff
@@ -625,6 +658,166 @@ interface class SoLoud {
...
+  ///
+  /// This method is a wrapper around [addAudioDataStreamU8].
+  void addAudioDataStreamS32(AudioSource source, Int32List audioChunk) {
+    addAudioDataStreamU8(source, audioChunk.buffer.asUint8List());
+  }
+
+  /// Set the end of the data stream.
+  ///
+  /// By setting the stream to be ended means that when playing it, it will stop
+  /// at that point (if the loop is not set).
```
filiph, line 781: What does it mean when the loop is set, though? Maybe add a paragraph saying what it means to call this function when `loop` is set.

src/active_sound.h
```diff
@@ -1,27 +1,34 @@
...
 
 /// The default number of concurrent voices - maximum number of "streams" - is 16,
 /// but this can be adjusted at runtime
-typedef struct ActiveSound
+struct ActiveSound
 {
     std::unique_ptr<SoLoud::AudioSource> sound;
     SoundType soundType;
-    std::vector<SoLoud::handle> handle;
+    std::vector<ActiveHandle> handle;
```
filiph, line 27: I'm starting to miss documentation here. Is this the list of sound handles that are currently pointing to (i.e. playing) this `ActiveSound`?

src/audiobuffer/audiobuffer.cpp
```diff
@@ -0,0 +1,252 @@
...
+			memset(aBuffer, 0, sizeof(float) * aSamplesToRead);
+
+			if (!mParent->dataIsEnded)
+			{
+				// The time the buffering is started.
+				// SoLoud::time currBufferTime = 1.0f * mOffset /
+				// 						  mParent->mPCMformat.bytesPerSample /
+				// 						  mParent->mChannels /
+				// 						  mParent->mPCMformat.sampleRate;
+				// std::cout << "CPP Buffering" << std::endl;
```
filiph, line 52: Should this still be here?

src/audiobuffer/audiobuffer.cpp
```diff
@@ -0,0 +1,252 @@
...
+			// So, if 1024 samples are requested from a stereo audio source, the first 1024 floats
+			// should be for the first channel, and the next 1024 samples should be for the second channel.
+			unsigned int i, j;
+			for (i = 0; i < samplesToRead; i++)
+			{
+				for (j = 0; j < mChannels; j++)
+				{
+					aBuffer[j * samplesToRead + i] = buffer[mOffset + i * mChannels + j];
+				}
+			}
```
filiph, line 73: [nit, probably premature optimization] Maybe, for data locality sake, it would be better to have the inner loop outside and the outer loop inside. So we go from byte to neighboring byte instead of jumping between distant channel data.

src/audiobuffer/audiobuffer.cpp
```diff
@@ -0,0 +1,252 @@
...
+			{
+				mParent->handle[i].bufferingTime = currBufferTime;
+				mThePlayer->setPause(mParent->handle[i].handle, true);
+				if (mOnBufferingCallback != nullptr)
+				{
+#ifdef __EMSCRIPTEN__
+						// Call the Dart callback stored on globalThis, if it exists.
+						EM_ASM({
+							// Compose the function name for this soundHash
+							var functionName = "dartOnBufferingCallback_" + $3;
```
filiph, line 190: Oh, here it is. It would be good to at least document where this is coming from, on both sides. Also add something like "If you change this here, you must also change the corresponding string on the {Dart/C++} side: path/to/file."

src/audiobuffer/buffer.h
```diff
@@ -0,0 +1,109 @@
+#ifndef CIRCULAR_BUFFER_H
+#define CIRCULAR_BUFFER_H
+
+// #ifndef _IS_WIN_
+// #include <algorithm>
```
filiph, line 5: Should this be here still?

src/audiobuffer/buffer.h
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    void setSizeInBytes(size_t newBytes)
+    {
+        maxBytes = newBytes;
+    }
+
+    // Overload for int8_t data, converting it to normalized float and adding its bytes to the buffer
+    // Return the number of floats written.
+    size_t addData(const int8_t* data, size_t numSamples) {
```
filiph, line 44: Do we really need all these overloads? These are basically internal methods, right? Maybe, to keep the API as minimal as possible, it would be better to just keep a single method?

src/audiobuffer/buffer.h
```diff
@@ -0,0 +1,109 @@
+#ifndef CIRCULAR_BUFFER_H
```
filiph, line 1: Is this a circular buffer?

src/enums.h
```diff
@@ -99,4 +105,24 @@ typedef enum FilterType
     PitchShiftFilter
 } FilterType_t;
 
+typedef enum BufferPcmType
```
filiph, line 108: Maybe add a comment about this enum also existing on the Dart side? And add a note about keeping the two in sync?

src/player.cpp
```diff
@@ -304,6 +305,72 @@ PlayerErrors Player::loadMem(
...
+    unsigned int hash,
+    const unsigned char *data,
+    unsigned int aDataLen)
+{
+    auto const s = findByHash(hash);
+
+    if (s == nullptr || s->soundType != TYPE_BUFFER_STREAM)
+        return soundHashNotFound;
+
+    return static_cast<SoLoud::BufferStream *>(s->sound.get())->addData(data, aDataLen);
```
filiph, line 349: [nit] Maybe check for this and raise an error or something if the sound isn't a buffer stream?

src/player.cpp
```diff
@@ -644,8 +717,13 @@ int Player::countAudioSource(unsigned int soundHash)
...
 
-    // if (s->soundType == TYPE_WAVSTREAM)
-    SoLoud::AudioSource *as = static_cast<SoLoud::WavStream *>(s->sound.get());
+    if (s->soundType == TYPE_WAVSTREAM)
+    {
+        SoLoud::AudioSource *as = static_cast<SoLoud::WavStream *>(s->sound.get());
+        return soloud.countAudioSource(*as);
+    }
+
+    SoLoud::AudioSource *as = static_cast<SoLoud::BufferStream *>(s->sound.get());
```
filiph, line 726: [nit] I think it makes sense now to `switch` over `s->soundType` instead of having if statements and the last possibility outside an if statement.

src/player.cpp
```diff
@@ -682,7 +760,7 @@ ActiveSound *Player::findByHandle(SoLoud::handle handle)
         int index = 0;
         while (index < (int)sounds[i].get()->handle.size())
         {
-            if (sounds[i].get()->handle[index] == handle)
+            if (sounds[i].get()->handle[index].handle == handle)
```
filiph, line 763: Oooh, a bug fix?

src/audiobuffer/audiobuffer.cpp
```diff
@@ -0,0 +1,252 @@
...
+			}
+		}
+
+
+		// data has been added to the buffer, but not all because reached its full capacity.
+		// So mark this stream as ended and no more data can be added.
+		if (bytesWritten < aDataLen / mPCMformat.bytesPerSample)
+		{
+			dataIsEnded = true;
+			return pcmBufferFullOrStreamEnded;
```
filiph, line 235: I'd prefer having two separate errors: one for full buffer, second for stream ended. This makes it much easier for the developer to debug problems, as they don't need to investigate 2 options instead of one.

filiph (changes requested):
I have some suggestions but this looks great, thank you!

example/lib/buffer_stream/generate.dart
```diff
@@ -0,0 +1,263 @@
...
+  final frequency = args['frequency'] as double;
+
+  // Sampling rate in Hz (samples per second)
+  final sampleRate = args['sampleRate'] as int;
+
+  // Duration of the audio in seconds
+  final duration = args['duration'] as double;
+
+  // Total number of samples needed for the given duration
+  final sampleCount = sampleRate * duration;
```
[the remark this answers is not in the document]
alnitak: done.

lib/src/bindings/bindings_player_web.dart
```diff
@@ -185,6 +189,78 @@ class FlutterSoLoudWeb extends FlutterSoLoud {
...
+      maxBufferSize,
+      bufferingTimeNeeds,
+      sampleRate,
+      channels,
+      pcmFormat,
+      // not used on C side. The callback is set below on the JS side. Setting
+      // this to 1 to tell C that we have a callback.
+      onBuffering == null ? 0 : 1,
+    );
+    final hash = wasmGetI32Value(hashPtr, '*');
```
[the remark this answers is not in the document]
alnitak: Don't remember where I read that `*` means `i32`. But better to explicit the type to `i32` as you noticed! (Also modified it on some other lines).
Done.

lib/src/bindings/bindings_player_web.dart
```diff
@@ -185,6 +189,78 @@ class FlutterSoLoudWeb extends FlutterSoLoud {
...
+    final soundHash = SoundHash(hash);
+    final ret = (error: PlayerErrors.values[result], soundHash: soundHash);
+    wasmFree(hashPtr);
+
+    if (onBuffering != null) {
+      // Convert Dart closure to a JS function. To the function name is added
+      // the hash of the sound to make it unique.
+      globalThis.setProperty(
+        'dartOnBufferingCallback_$hash'.toJS,
```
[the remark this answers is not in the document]
alnitak: Here a new JS function is created which will call the Dart callback provided in the `setBufferStream():onBuffering` parameter closure.
I've added comments here and also on the C++ side where you noticed this in the conversation below.
Done.

lib/src/enums.dart
```diff
@@ -64,8 +64,12 @@ enum PlayerErrors {
   /// Error getting filter parameter.
   filterParameterGetError(18),
 
+  /// Trying to add PCM data but the buffer is full or stream buffer has been
+  /// set to be ended.
+  pcmBufferFullOrStreamEnded(19),
+
   /// No playback devices were found.
-  noPlaybackDevicesFound(19);
+  noPlaybackDevicesFound(20);
```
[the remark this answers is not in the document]
alnitak: Don't remember why this happened.
Rolling back `noPlaybackDevicesFound` to 19 and the new `pcmBufferFullOrStreamEnded` to 20.

lib/src/soloud.dart
alnitak: done.

lib/src/soloud.dart
alnitak: Yep, just that.
Done.

lib/src/soloud.dart
alnitak: Added a short example here. Also a reference to the `example/buffer_stream/generate.dart` example (where I forgot to comment on the purpose of that :) ) which demonstrates how to generate 3 kinds of audio data using `compute`.

lib/src/soloud.dart
alnitak: Good point. I was concerned by this because other typed data are not foreseen ie `Int32x4List`. But anyway, they can be casted to Uint8List, so it's good to have only one method.
Done.

lib/src/soloud.dart
alnitak: Yes right, my explanation is bad. Corrected to:
```
  /// By setting the stream to be ended means that when playing it, it can
  /// handle the stop event when it reaches the end of the data stream or can
  /// be looped if the looping is enabled.
```
Done.

src/active_sound.h
alnitak: You are correct. That `handle` is still a list of handles currently active in the `AudioSource`, but I had to add a parameter for each handle to know if/when the handle has reached the end of the data to catch the buffering event. This is managed only in the new `audiobuffer.cpp` and it doesn't affect anything else.
Added a comment here.

src/audiobuffer/audiobuffer.cpp
alnitak: Done.

src/audiobuffer/audiobuffer.cpp
alnitak: Nice catch!
Done.

src/audiobuffer/audiobuffer.cpp
alnitak: comments added on both sides.
Done.

src/audiobuffer/buffer.h
alnitak: Removed.

src/audiobuffer/buffer.h
alnitak: Done.

src/enums.h
alnitak: Done.

src/player.cpp
alnitak: Done.

src/player.cpp
alnitak: Done.

src/player.cpp
alnitak: nono! That's the new struct `ActiveHandle` which contains the `handle` and the `bufferingTime` :)
Used in `ActiveSound` as a list of active handles.

src/audiobuffer/audiobuffer.cpp
alnitak: Done.

filiph:
> I thought it would be interesting to have a new method to reset the BufferStream. Something like `void resetBufferStream(AudioSource source)` which removes the ended flag and frees up the old audio data allowing the developer to add new audio data from scratch?

To be honest, I'm not the intended audience (I don't foresee needing buffer stream) so I'm not the right person to ask. Maybe the folks who're asking for the feature might tell you? Otherwise, in general, I'm a proponent of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

prateekmedia:
@alnitak The resetBufferStream method can be useful when audio stream needs to be interuppted due to any reason and new audio stream needs to feed.

alnitak:
Thanks @prateekmedia,

I was thinking of the `resetBufferStream` because I was trying to figure out how to do a `walky_talky` like app or any other workflow like this:
1. setup an AudioSource with `setBufferStream`
2. receive data from the net and add them to the AudioSource while playing it
3. when data is ended call `resetBufferStream` and go back to `2` instead of disposing of the AudioSource and going to `1`

Anyway, let's flush this PR and shortly we will think better about what will be useful to add.

keyur2maru:
@alnitak thank you for building this library. Are there any plans of implementing `resetBufferStream` functionality? I was looking into using this for my voice enabled application which supports interruptions to the streaming audio based on detected voice activity and then it fetches new streaming audio data. Having an option to call `resetBufferStream` would allow me to target this use case precisely.

alnitak:
@keyur2maru the `resetBufferStream` is a reasonable addition! 

I am opening a new issue to reference and to take track of this.

PR #149, opened 2024-11-11
Limiter and Compressor filters

alnitak:
## Description

Added Limiter and Compressor filters.

Happens that playing multiple sounds at the same time can create noise and output distortion. The limiter or the compressor filter can be used to reduce this behavior.
I think the compressor is better suited for games.

Both come with an example in the `example/lib/filters` dir. Both examples play the same audio at the same time to gain a high and distorted volume. The sliders are used to arrange the filter parameters and find which values are the best to use.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

I tried to do the limiter filter, but I wasn't happy with the results it can have for your use case. So I've also made a compressor filter that may work better for you. I just pushed a PR but starting this I merged the `websocket` into this new PR and I think you will also see the other changes. Anyway, there is not much Dart code in that to review. So I think it is better first to review the `websocket` (wrong branch name anyway :) ) and then just try the filters PR.


filiph:
Thanks for doing this! If you don't mind, I'll wait for #148 to be merged and this one to be rebased so that I can have a cleaner view of the changes here.

PR #178, opened 2025-02-09
Use native controller call to check ready state.

### LINKED ISSUE #177: fix: isInitialized returning false on MacOS after engine starts with no error
**Description**

When initializing the engine on MacOS, the function `getVisualizationEnabled()` is called. 

This function immediately checks `if (!isInitialized)` and throws an error with message if not. 

The problem is, the engine has been initialized and the value of `_isInitialized` is true.

The engine started with no error, and `_isinitialized` is set a few lines above the call to `getVisualizationEnabled()`. Maybe we can use this dart variable instead of value of `mInited`.

I believe at this point, the engine should be returning that it is initialized in order to proceed. When examining the code, function `_controller.soLoudFFI.isInited()` is what is controlling the bool `isInitialized`.

`isInited()` is checking `bindings.cpp` `isInited` which returns 0 if the player is Null or `mInited` is 0.
At this point in the cycle, I believe the player might still be null, or engine has not signaled ready. `mInited` would still be false if the engine is still initializing. Seems like a race condition. Either way, the `await` on `soloud.init()` is releasing too soon. 

**Steps To Reproduce**

1. Boot application on MacOS
2. call soloud.init()
3. engine reports success from internal init method (SoLoud_NoError)
4. await from call is returned with fail state, apparently `mInited` is still false.
5. application returns error state while proceeding to `_isVisualizationEnabled` 364 (3.0.0-pre.0)
6. See error**Expected Behavior**

Audio Engine should initialize and return `isInitialized = true`

Feel free to assign me unless you have a quick fix.

themailman05:
Closes #177

## Description

getVisualizationEnabled() relied on a complex path to check ready state of engine during dart init procedure. Instead, we can call the native binding to the C function which should expose our engine initialization state truthfully.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/soloud.dart
+++ b/lib/src/soloud.dart
@@ -361,7 +361,7 @@ interface class SoLoud {
       /// get the visualization flag from the player on C side.
       /// Eventually we can set this as a parameter during the
       /// initialization with some other parameters like `sampleRate`
-      _isVisualizationEnabled = getVisualizationEnabled();
+      _isVisualizationEnabled = _controller.soLoudFFI.getVisualizationEnabled();
 
       // Initialize [SoLoudLoader]
       _loader.automaticCleanup = automaticCleanup;


alnitak:
Thanks @themailman05,

I will test this too and merge it as soon as possible even if it's not fully clear to me why you get that misbehavior. However, this change can only do good because it is an optimization.

> I will write a good test too. I don't have all supported platforms running yet with my system, so in no rush to ship to main. 

you could try with `example/tests/tests.dart` and run `testAsynchronousDeinit` and `testSynchronousDeinit`. These are the most relevant tests to try for the `init` method.

Issue #104, opened 2024-07-09
negative interaction with audio_session


lukepighetti:
https://github.com/ryanheise/audio_session/issues/138

alnitak:
Hey @lukepighetti,

I read **ryanheise** reply. Unfortunately, I don't have a Mac to try. What I can say is that `loadAsset()` does nothing else than read using `rootbundle` or give the file directly to the native if it is already cached.

In the `flutter_soloud` version you are using, an `Isolate` is used to perform the native call. This could be the only reason that come to my mind that could cause something out of the ordinary. The `main` branch got rid of the Isolate and maybe worth to try it. It's not yet been published, so you should link this repo in your `pubspec.yaml` like:
```
  flutter_soloud:
    git:
      url: https://github.com/alnitak/flutter_soloud.git
      # Add this if you want to try with the incoming web support.
      # path: web
```
Also, have you tried to load the asset before calling `_session.setActive(true)`?

sbauly:
I’m also experiencing issues with SoLoud’s interaction with audio_session, although I haven’t experienced the same issue as @lukepighetti.

It seems that the way SoLoud is handling interruptions internally could be interfering with attempts to use audio_session.

I made a quick example app that highlights the issue on iOS. The app uses one AudioSession instance with the exact same implementation for just_audio and SoLoud.

https://github.com/user-attachments/assets/fe7e8012-121c-4873-ab8a-e1f00bbe56af

You can see from the video:

1. When an interruption occurs in SoLoud, the audio will not resume. You have to call `SoLoud.instance.init()` again and reload the asset before you can call `play()` successfully. This is not ideal, as it adds a lot of code and also means that the audio must be restarted again from the very beginning of the track after an interruption, losing the position of the audio that was being played.

2. When an interruption occurs in just_audio, the session correctly loses and regains focus as you would expect, maintaining the position of the audio.

Ideally, SoLoud would leave all interruption logic to be handled in audio_session, which would allow for one centralized place for session logic and handling of more complex scenarios. I’ve been using the main branch as suggested by @alnitak, and also tried using a local fork of SoLoud, commenting out some of the interruption logic I could find like `ma_device_notification_type_interruption_began` - but I haven’t been able to find a fix yet. This only appears to be a problem on iOS.

alnitak:
Hi @sbauly, thank you very much for your findings and your time.

[...]

PR #183, opened 2025-02-18
Android, iOS and MacOS audio focus

alnitak:
## Description

This would add new audio focus interruptions and management to the plugin.

The question is: is it worth intercepting the interrupt events or instead it would be better to use audio_session and try to disable in miniaudio session?
I saw that maybe it is possible [using this](https://github.com/alnitak/flutter_soloud/blob/936def7dac5337427f4cd65d03ca23886bd81116/src/soloud/src/backend/miniaudio/miniaudio.h#L8315)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


sbauly:
Hey @alnitak,

I have been testing this PR some more and made some observations:

- The app will crash if you attempt to init & play the asset _before_ calling `SoLoud.instance.interruptions.stateChangedEvents.listen()`, but it works fine if you call listen() first.

- The `ended` and `endedResume` methods appear to behave the same way currently - both of them will resume audio immediately after an interruption ends.

As mentioned in #104, I think the ideal scenario would be to decouple the interruption logic from SoLoud in favor of using the `audio_session` package to handle different use cases.

I've looked into this, but have been a little intimidated by the almost 100,000 lines of code within `miniaudio.h` 😅

Please let me know if I can continue to help with testing this PR in any way on iOS and/or MacOS, or anything else I can help with in general.

Thanks again!

alnitak:
Hey @sbauly,

Thanks for your testing! I agree with you. Using audio_session to handle interruptions is a better approach. Adding this workflow for all the platforms and maintaining it could be problematic for me and maybe out-of-scope of flutter_soloud.

I close this PR and leave the `interruptions` branch alive for a while.

Thanks a lot for all!

sbauly:
Thanks @alnitak!

Something that this branch did achieve successfully (that I think is worth pursuing) is that it stopped the SoLoud engine from deinitializing completely each time an interruption event occurred on iOS. It would be great if we could get this functionality working on the main branch, as whilst `audio_session` does _mostly_ work with SoLoud currently, the deinit/reinit fix on iOS does have limitations.

Issue #191, opened 2025-02-28
fix: macOS build fails in latest version


filiph:
**Description**

I hope to investigate this further, but after upgrading to `flutter_soloud` 3.0.2 (from pre-3 version), I get the following build fail on macOS:

```
ld: symbol(s) not found for architecture x86_64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

**Steps To Reproduce**

I haven't tried with a smaller codebase yet, so here's the "largest" reproducible sample:

1. Clone the GIANT ROBOT GAME (you have access) at commit 2daa91296f65f10ed162a96997aefcdb645f4474 (currently latest)
2. Run `flutter build macos` on a mac

**Expected Behavior**

App builds.

**Additional Context**

Here's the full output, including warnings:

```
Cleaning Xcode workspace...                                         3.3s
Deleting build...                                                1,889ms
Deleting .dart_tool...                                             135ms
Deleting ephemeral...                                                1ms
Deleting ephemeral...                                                2ms
Deleting ephemeral...                                                1ms
Deleting .flutter-plugins-dependencies...                            0ms
Deleting .flutter-plugins...                                         0ms
Resolving dependencies... 
Downloading packages... 
  async 2.12.0 (2.13.0 available)
  fake_async 1.3.2 (1.3.3 available)
  leak_tracker 10.0.8 (10.0.9 available)
  material_color_utilities 0.11.1 (0.12.0 available)
  vm_service 14.3.1 (15.0.0 available)
  webdriver 3.0.4 (3.1.0 available)
Got dependencies!
6 packages have newer versions incompatible with dependency constraints.
Try `flutter pub outdated` for more information.

Running pod install...                                           1,921ms
--- xcodebuild: WARNING: Using the first of multiple matching destinations:
{ platform:macOS, arch:arm64, id:00008103-00141980342A001E, name:My Mac }
{ platform:macOS, arch:x86_64, id:00008103-00141980342A001E, name:My Mac }
:70:23: warning: '&' within '|' [-Wbitwise-op-parentheses]
   70 |         return (value & 127 | i << 7) + 128;
      |                 ~~~~~~^~~~~ ~
   70 |         return (value & 127 | i << 7) + 128;
      |                       ^
      |                 (          )
[...]

Issue #192, opened 2025-03-01
fix: Failed to load dynamic library 'flutter_soloud_plugin.dll' on Windows


filiph:
**Description**

After upgrading to `flutter_soloud` 3.0+ (from 2.1.7), I got the following error report through Sentry. I'm not seeing this on my own Windows computer.

```
ArgumentError: Invalid argument(s): Failed to load dynamic library 'flutter_soloud_plugin.dll': The specified module could not be found.

 (error code: 126)
  #0      _open (dart:ffi-patch/ffi_dynamic_library_patch.dart:11)
  #1      new DynamicLibrary.open (dart:ffi-patch/ffi_dynamic_library_patch.dart:22)
  #2      new SoLoudController._ (package:flutter_soloud/src/bindings/soloud_controller_ffi.dart:19)
  #3      new SoLoudController (package:flutter_soloud/src/bindings/soloud_controller_ffi.dart:10)
  #4      new SoLoud._ (package:flutter_soloud/src/soloud.dart:74)
  #5      SoLoud.instance (package:flutter_soloud/src/soloud.dart:166)
  #6      DefaultAudioController._initializeSoLoud (package:giant_robot/audio/src/soloud_audio_controller.dart)
```

**Steps To Reproduce**

None so far.

**Expected Behavior**

Library initializes like normal.

**Screenshots**

![Image](https://github.com/user-attachments/assets/2e1c200e-863d-4639-98a3-7b6f2e99972b)

![Image](https://github.com/user-attachments/assets/776f02f9-36af-487a-9516-65d3e665bc23)

quadbyte:
I'm having the same issue.  It seems its related to "flutter_soloud_plugin.dll"  being dependent on "opus.dll"
And "opus.dll" requires  VCRUNTIME140D.dll (debug version)  and (ucrtbased.dll) which are not installed on a normal windows system, they usually get installed with Visual Studio.

So in my case, Flutter SoLoud is working ok on my dev machines, but never inits on any of my clients computer.

Inspected the dependencies with https://github.com/lucasg/Dependencies

![Image](https://github.com/user-attachments/assets/68fe5a32-c888-40f0-82c5-046d5061883a)

alnitak:
Thanks @quadbyte, it would have taken me a while to find out!

[...]

PR #194, opened 2025-03-02
fix: libopus and libogg build issue on macOS

filiph:
## Description

This is a workaround for https://github.com/alnitak/flutter_soloud/issues/191 (not a fix, see below).

It will correctly avoid building apps with libopus + libogg unless the development mac has _both_ libraries in _both_ architectures (`x86_64` and `arm64`).

When both these libraries in both architectures are present on the development machine (see [here](https://github.com/alnitak/flutter_soloud/issues/191#issuecomment-2692671697) for a how to), only then will the `LIBOPUS_OGG_AVAILABLE` flag be set. The podspec file then correctly points to the correct library paths (`/opt/homebrew/lib` for arm64, `/usr/local/lib` for x86_64), so the build will succeed.

While this fixes the build, it doesn't fix the issue, imho. We can't assume that the _user_ has the libraries installed on their machine. And if they don't, the app will break at runtime.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


macos/flutter_soloud.podspec
```diff
@@ -43,9 +50,11 @@ Flutter audio plugin using SoLoud library and FFI
     'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386',
     "CLANG_CXX_LANGUAGE_STANDARD" => "c++17",
     "CLANG_CXX_LIBRARY" => "libc++",
-    'OTHER_LDFLAGS' => has_opus_ogg ? '-L/usr/local/lib -L/opt/homebrew/lib -logg -lopus' : '',
+    'OTHER_LDFLAGS[arch=arm64]' => has_opus_ogg ? '-L/opt/homebrew/lib -logg -lopus' : '',
+    'OTHER_LDFLAGS[arch=x86_64]' => has_opus_ogg ? '-L/usr/local/lib -logg -lopus' : '',
     'OTHER_CFLAGS' => '-msse -msse2 -msse3 -msse4.1 -O3 -ffast-math -flto',
-    'OTHER_CPLUSPLUSFLAGS' => '-msse -msse2 -msse3 -msse4.1 -O3 -ffast-math -flto'
+    'OTHER_CPLUSPLUSFLAGS' => '-msse -msse2 -msse3 -msse4.1 -O3 -ffast-math -flto',
+    'VALID_ARCHS' => 'x86_64 arm64',
```
filiph, line 57: Not sure if this is needed but kept it in.

alnitak: I'm not sure either, but it certainly won't hurt.

alnitak:
Thanks Filip!

I am working on the embedding solution we talked about and since I don't know how much it will take, this workaround is very welcomed! I think I am already at a good point but lets merge this waiting for the fix.

Do you think that releasing on pub.dev could wait a few days (1 maybe 2)?

filiph:
> Do you think that releasing on pub.dev could wait a few days (1 maybe 2)?

Of course, it can wait. There probably aren't that many people using `flutter_soloud` for macOS, the issue only happens when you have `libopus` and `libogg` installed on your dev mechine (which isn't rare but it's not 100% of machines), and this issue has been out since 3.0 anyway (i.e. over 2 weeks). So 1 or 2 days is totally fine. (As for me, I'm on pre-3.0 version at least until this and #192 are resolved.)

PR #205, opened 2025-03-19
added  a check for max voices when calling play

### LINKED ISSUE #204: fix: Sounds seemingly "backed up in a queue" when playing too many at once
**Description**

Playing many sounds at once has a strange effect where some of the sounds seemingly play long after they were played.

See the following video:

https://github.com/user-attachments/assets/33ac0090-1262-4d45-a185-db75011c1a4e

Notice how after about 0:07, sounds start playing again. None of these are actually supposed to play at that time.

**Steps To Reproduce**

1. Get Giant Robot Game source (I know you have access, @alnitak) and run it (the video above is from Windows)
2. Click Custom mission, get as many missiles as possible
3. Start mission
4. Move cursor over map and hit `Q` twenty times in quick succession

**NOTE:** I'm pretty sure you can reproduce this just by playing a bunch of sounds in quick succession in any codebase. But I haven't taken the time to try it so I'm listing the one reproduction I can confirm at this point.

**Expected Behavior**

Audio mayhem, but it stops after a few seconds.

**Actual Behavior**

Weird "echoes" — probably late triggers?

**Additional Context**

I don't think I'm doing anything fancy with playing SFX. You can check `lib/audio/src/soloud_audio_controller.dart`, line 338.

alnitak:
## Description

Check if by playing a new sound will exceed the maximum number of voice count. If true, then check if the `AudioSource` has other instances playing. If true remove the first and play the new one.
If there are no other instances playing, this sound cannot be played and it will be dropped. In this latter case an error will be returned which is triggered not as an exception, but just as a log warning since it is a non-blocking situation.

fixes #204

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


alnitak:
This PR tries to play the sound even if the max active voice count has been reached by stopping the oldest sound of the same `AudioSource` instance.

This happens for example in a game when using a firing sound multiple times (more than 16). When this happens, the oldest sound is stopped in favor of the new one.

But if the sound has no other instances, the play will be dropped and the solution is only to raise the max active voice number.

In both circumstances, it will be displayed as just a log warning informing the dev.

lib/src/enums.dart
```diff
@@ -163,6 +167,13 @@ enum PlayerErrors {
       case PlayerErrors.bufferStreamCanBePlayedOnlyOnce:
         return 'BufferStream can be played only once when using '
             '`BufferingType.release` buffer type!';
+      case PlayerErrors.maxActiveVoiceCountReached:
+        final max = SoLoud.instance.getMaxActiveVoiceCount();
```
filiph, line 171: Probably not a big deal, but maybe calling this (FFI) function from an error stringifier might be a bit dangerous. It could lead to some kind of "error while trying to create error message". I'd personally prefer just say "maximum reached", without the concrete number. Plus, you won't have to add the circular dependency back to the main `flutter_soloud.dart` file.

src/player.cpp
```diff
@@ -536,6 +536,20 @@ PlayerErrors Player::play(
         return bufferStreamCanBePlayedOnlyOnce;
     }
 
+    // Check if by playing this sound will exceed the maximum number of voice count. If true, then
+    // check if [soudHash] has other instances playing. If true remove the first and play the new one.
+    // If there are no other instances playing, this sound cannot be played and return an error.
+    // Issue https://github.com/alnitak/flutter_soloud/issues/204
+    if (getActiveVoiceCount() >= getMaxActiveVoiceCount())
```
filiph, line 543: Just checking: is either of these functions "expensive"? Like, is it just checking a number, or is it counting an array length? Or, worse, trying to walk some more complicated collection?

src/player.cpp
```diff
@@ -536,6 +536,20 @@ PlayerErrors Player::play(
...
+    {
+        if (sound->handle.size() > 0)
+        {
+            stop(sound->handle[0].handle);
```
filiph, line 547: Are we absolutely certain that this is safe? Is it possible the sound is already stopped? Or there's a null somewhere in here?

Also, are we sure that `handle[0]` is the oldest handle?

filiph: Also, are we sure that stopping here will immediately clear the voice for the sound to be played later on line 554?

filiph: If any of this isn't 100% crystal clear, or if there's a possibility this will be a maintainability issue, I'd strongly suggest we just drop the new sound and return the error, not trying to play it. Again, we're talking about an edge case. We don't need to go out of our way to make this work, and I worry this will be a bug magnet.

src/player.cpp
```diff
@@ -1007,6 +1021,20 @@ PlayerErrors Player::play3d(
...
```
filiph, line 1032: Same here as above.

filiph (commented):
Wow, thanks for the really quick turnaround! I have some questions inline.

lib/src/enums.dart
alnitak: Done.

alnitak:
> Does it overcount or undercount?

it could overcount, counting also the distant voices, hence those with a very low volume (0.001 IIRC) or paused.

filiph:
Maybe I'm overthinking this, but I'm not a big fan of having two "versions" of the same number. I think it could lead to subtle bugs (e.g. I "know" there's enough room to play another sound right now but it won't play).

Could we just make the new (internal) version the only version? Just drop the heuristic of the current (public) version of getActiveVoiceCount?

alnitak:
> Could we just make the new (internal) version the only version? Just drop the heuristic of the current (public) version of getActiveVoiceCount?

Oh, I see! If you tried to tell me before, I'm sorry but I didn't understand! Yes, that's a good point.

Fixed.

filiph:
I see you're still keeping the original version of `getActiveVoiceCount` around. Is that needed? I was hoping we could just get rid of that code completely, or at least make it `deprecated` or `experimental`. Or do you think some people are depending on the current heuristic in the `getActiveVoiceCount` functino?

alnitak:
`Player::getActiveVoiceCount()` can be safely removed in favor of `Player::getActiveVoiceCount_internal()`. I left it on C side just because is part of SoLoud C++ lib, but it was not accessible from Dart.

I removed it.

filiph:
Oh, I didn't realize (or forgot) it was a soloud c++ method.

Anyway, I'm happy if you're happy.

filiph (approved):

alnitak:
Version 3.1.1 is out on pub.dev!

PR #215, opened 2025-03-29
log maxActiveVoiceCountReached with Level.INFO #212

### LINKED ISSUE #212: fix: Reaching maxActiveVoiceCount should not raise a severe log event
**Description**

I guess I missed this in the previous PR review. It seems that the new version raises a severe log every time the max active voices count is reached. This is not a good idea.

**Steps To Reproduce**

1. Play lots of sounds at once.

**Expected Behavior**

Sounds over "budget" are dropped. An info-level message is generated.

**Screenshots**

Image

**Additional Context**

I really don't think that reaching the maximum voices is a SEVERE-type error. At most, it's a warning, but it's probably just an INFO-level event. In busy games, it is simply going to happen from time to time unless the developer sets some absurd limit.

alnitak:
## Description

Fixes #212

`PlayerErrors.maxActiveVoiceCountReached` should log with `Level.INFO` instead of `Level.SEVERE`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/lib/src/soloud.dart
+++ b/lib/src/soloud.dart
@@ -2559,7 +2559,8 @@ interface class SoLoud {
   }
 
   /// Utility method that logs a [Level.SEVERE] message if [playerError]
-  /// is anything other than [PlayerErrors.noError].
+  /// is anything other than [PlayerErrors.noError] or [Level.INFO] if
+  /// the error is [PlayerErrors.maxActiveVoiceCountReached].
   ///
   /// Optionally takes a [from] string, so that it can construct messages
   /// with more context:
@@ -2578,7 +2579,8 @@ interface class SoLoud {
       return;
     }
 
-    if (!_log.isLoggable(Level.SEVERE)) {
+    if (!_log.isLoggable(Level.SEVERE) &&
+        playerError != PlayerErrors.maxActiveVoiceCountReached) {
       // Do not do extra work if the logger isn't listening.
       return;
     }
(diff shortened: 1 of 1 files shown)


lib/src/soloud.dart
```diff
@@ -2578,7 +2579,8 @@ interface class SoLoud {
       return;
     }
 
-    if (!_log.isLoggable(Level.SEVERE)) {
+    if (!_log.isLoggable(Level.SEVERE) &&
+        playerError != PlayerErrors.maxActiveVoiceCountReached) {
       // Do not do extra work if the logger isn't listening.
       return;
     }
```
filiph, line 2586:
```suggestion
    // Do not do extra work if the logger isn't listening
    // to the appropriate level.
    final logLevel = playerError == PlayerErrors.maxActiveVoiceCountReached
        ? Level.INFO
        : Level.SEVERE;

    if (!_log.isLoggable(logLevel)) {
      return;
    }
```

I got lost in this logic so I'm suggesting this change. It goes with the suggestion below.

lib/src/soloud.dart
```diff
@@ -2588,6 +2590,10 @@ interface class SoLoud {
       strBuf.write('$from: ');
     }
     strBuf.write(playerError.toString());
-    _log.severe(strBuf.toString());
+    if (playerError == PlayerErrors.maxActiveVoiceCountReached) {
+      _log.info(strBuf.toString());
+    } else {
+      _log.severe(strBuf.toString());
+    }
```
filiph, line 2597:
```suggestion
    _log.log(logLevel, strBuf.toString());
```

filiph (approved):
LGTM with an optional improvement.

Issue #226, opened 2025-04-18
fix: `listPlaybackDevices` fails to retrieve devices when the device prefix contains Chinese characters


WHYBBE:
**Description**

When the prefix of your device contains non - ASCII characters, the list returned by `listPlaybackDevices` will not include them. In particular, when all of your devices are in this situation, `listPlaybackDevices` will return an empty list.

**Steps To Reproduce**

1. You need to have an audio device that does not start with ASCII characters.
2. Try the output_device in the example.
3. You will see error.

**Expected Behavior**

The results returned by `listPlaybackDevices` contain devices with non - ASCII prefixes.

**Screenshots**

![Image](https://github.com/user-attachments/assets/c36ef96d-902a-4609-b337-aefad6800ef9)

alnitak:
Thank you @WHYBBE!

Your PR has been merged.

PR #227, opened 2025-04-18
fix `listPlaybackDevices` fails to retrieve devices when the device prefix contains Chinese characters

WHYBBE:
UTF-8 or other character sets may use the high - order 1 of ASCII (8 - bit). When the type of d[i].name[n] is unsigned char and the value is negative, it will be considered to contain special characters and thus filtered out. However, they are legitimate audio devices.

## Description

related to #226

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/bindings.cpp
+++ b/src/bindings.cpp
@@ -244,7 +244,7 @@ extern "C"
             /// to be sure.
             for (int n = 0; n < 5; n++)
             {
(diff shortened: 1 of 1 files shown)


alnitak:
@WHYBBE  thanks a lot!

Issue #249, opened 2025-05-25
feat: Audio Session Integration and Stream Type Control


hoalarious:
**Description**

Currently, flutter_soloud does not integrate with Flutter's audio session management system, causing it to ignore AndroidAudioAttributes configurations set via the audio_session package. This creates inconsistent behavior compared to other Flutter audio plugins and prevents proper audio stream routing on Android. 

On Android audio played by flutter_soloud can only be altered by changing the media volume. Not in call or notification volumes.

**Requirements**

- [ ] Audio Session Integration: Read and respect Flutter's audio session configuration from the audio_session package
- [ ] AndroidAudioUsage Support: Honor Android audio usage types: USAGE_MEDIA, USAGE_VOICE_COMMUNICATION, USAGE_NOTIFICATION, etc.

alnitak:
Hi @hoalarious,

in the new 3.2.0 version, you are able to use a 3rd party plugin like [audio_session](https://pub.dev/packages/audio_session) to manage audio focus.

Feel free to reopen this issue for any related problems! Thanks.

Issue #245, opened 2025-05-15
fix: No sound after showing rewarded ads with Admob on iOS


HasancanCakicioglu:
**Description**

The same code works fine on Android, but after showing the admob rewarded ad on iOS and after the user watches and closes the ad, there is no sound. There is no problem in the simulator, but there is a problem on real devices. I tried iPhone 14 Pro Max iOS 18.4.1 and there was no sound. It cannot be started afterwards.

alnitak:
This issue seems related to #183, which I put in stand-by. 

When a rewarded ad uses audio, flutter_soloud will lose the focus. I hadn't thought about this problem, so that PR becomes more important to achieve! But maybe [audio_session](https://pub.dev/packages/audio_session) package can be used to work around the problem  #104

alnitak:
Hi @HasancanCakicioglu,

in the new 3.2.0 version, you are able to use a 3rd party plugin like [audio_session](https://pub.dev/packages/audio_session) to manage audio focus.

Feel free to reopen this issue for any related problems! Thanks.

orkun1675:
I am able to replicate this on `flutter_soloud: 3.3.1` and Flutter 3.32.8.

Here is a sample project: https://github.com/orkun1675/flutter_soloud_test

Steps to reproduce:
- Run the sample app
- Turn on background music and sfx switches
- Background music begins, the "Play SFX" button works
- Show an AdMob ad, during the add notice that the SoLoud audio is correctly paused
- Close the ad
- The SoLoud audio is broken, can no longer be played

Things I've tried, without luck:
- Installing and enabling audio_session at app start & after AdMob ad is dismissed
- Re-initializing SoLoud after AdMob ad is dimissed

Flutter logs:
```
Launching lib/main.dart on Orkun’s iPhone (wireless) in debug mode...
Automatically signing iOS for device deployment using specified development team in Xcode project: WQ65PJ26MP
Xcode build done.                                            8.6s
You may be prompted to give access to control Xcode. Flutter uses Xcode to run your app. If access is not allowed, you can change this through your Settings > Privacy & Security > Automation.
Connecting to VM Service at ws://127.0.0.1:59851/NHtsPSvdr10=/ws
Connected to the VM Service.
[flutter_soloud.SoLoudLoader_io] Temporary directory initialized at /var/mobile/Containers/Data/Application/E042BEAD-9387-43E5-8127-CC1908BD322F/Library/Caches/SoLoudLoader-Temp-Files
[...]

PR #257, opened 2025-06-23
feat: trying to disable audio context

### LINKED ISSUE #104: negative interaction with audio_session

alnitak:
## Description

This PR disables the audio context and device routing of miniaudio, allowing developers to use a third-party plugin such as [audio_session](https://pub.dev/packages/audio_session) instead.

A new example has been added: `example/lib/audio_context/audio_context.dart`.

This has been tested and is working on Android. For example, after running the app and starting a YouTube video, the music is paused as expected. When an even arrives:

- switching to the YouTube app and playing a video triggers an `AudioInterruptionType.unknown` event with `event.begin=true`, but no corresponding `event.begin=false`.
- receiving a notification sound or a phone call triggers an `AudioInterruptionType.duck` event with `event.begin=true`.

I wasn't able to trigger `AudioInterruptionType.pause`. YouTube Music, YouTube Video, and other music apps seem to only trigger `AudioInterruptionType.unknown`.

Unfortunately, I don’t have access to a real iOS device to test this. If anyone is able to try it out on iOS, I’d greatly appreciate your feedback.

For those who wish to try this out, the core part to look at is in [miniaudio_init](https://github.com/alnitak/flutter_soloud/blob/c36e8dde41385f3fab78db89e2a8f00d0035d782/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp#L127).

Aiming to fix #104, #245 and #249

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


MinseokKang003:
https://github.com/user-attachments/assets/bc434009-b132-4547-a27b-fdc7fe77837e

https://github.com/user-attachments/assets/9f8911be-0484-46a1-a29f-67ae99e10b95

Hey @alnitak thanks for the clarification.
I've tested out two cases on my iPhone 13 Pro:
[play asset] -> [Interruption] -> [play asset] -> [unpause]
[play asset] -> [Interruption] -> [unpause]

But in both cases the audio does not regain after interruption.
Let me know if I need to do the other way.

alnitak:
I have pushed a commit that may solve this. I think you will need to `flutter clean` your project.

On iOS, when `unknown` interruption is triggered, the device is stopped (not uninitialized). I have forced it to start again. This was working in my older tests in another PR, but I am unsure if it will now. I have also added `await session.setActive(true);` in the `unpause` button.

So, you can now try again to:
[play asset] -> [Interruption] -> [unpause]

Please, can you also try to check the duck event?
- [play asset]
- [notification sound] the asset volume should duck (the volume should be lowered until the notification sound ends).

Thanks!

MinseokKang003:
https://github.com/user-attachments/assets/10c4a77a-82d1-4121-8d76-526c9f3b7410

I've run `flutter clean` and tested it again.
Still doesn't work, but if you see the recording I found when I background my app the audio starts playing again.
Maybe the interruption ended notification is not being called at the right time.

Logs:
```
flutter: audio_context: interruption begin: true
flutter: audio_context: interruption type: AudioInterruptionType.unknown
[App goes background]
flutter: audio_context: interruption begin: false
flutter: audio_context: interruption type: AudioInterruptionType.unknown
```

I'll try the duck event soon.
Thanks!

alnitak:
I'm really sorry I didn't get the chance to try it out myself. Anyway, looking at your recording, I think the audio engine has not been disposed, hence it doesn't need to be initialized again as before this PR.

When you press the `unpause` button, it seems that the play starts where it was interrupted by the interruption. This made me think that the audio context is not the same as before. Maybe worth trying adding `await session.setActive(false);` after the `soloud.setPause(soundHandle!, true);` in the `event.begin` of the `unknown` event.

I am not sure, and I am not very familiar with audio_session and how it works, but maybe now we have to look for something on the Flutter side, like the app life cycle, where, for example, try putting an `await session.setActive(true);` in the app resumed.

Maybe @adventureisyou or @sbauly can give us some insights?

MinseokKang003:
@alnitak I'm not familiar with audio systems at all but as I understand, the problem is `ma_device_start` not being called properly on audio playback, if `ma_device_stop` has been called by miniaudio's `AVAudioSessionInterruptionTypeBegan` event.

The problem on iOS is that `AVAudioSessionInterruptionTypeEnded` event is not guaranteed to be called, so we can't rely on that to call `ma_device_start`.

It may work if there is some way we could check, every time on `soloud.play` or `soloud.setPause`, if the device has been stopped and if so, we call `ma_device_start` explicitly.

For my use case the PR's current state works great because I'm using SoLoud for sound effects which mix with other audios.   It's great that I'm given explicit control over AVAudioSession using `audio_session`, which I'm using like:
```dart
await session.configure(
  const AudioSessionConfiguration(
    avAudioSessionCategory: AVAudioSessionCategory.playback,
    avAudioSessionCategoryOptions:
        AVAudioSessionCategoryOptions.mixWithOthers,
  ),
);
```

Please feel free to let me know if you need more tests.

MinseokKang003:
@alnitak I've opened #260 which implements my suggestion in my last comment. Tested on iPhone 13 Pro. Feel free to check! (I don't have much experience in posting PRs so sorry if this is the wrong way).

alnitak:
@MinseokKang003 I have merged your PR and updated the example a bit.

Could you please confirm that running the example and doing what is described in the first post under the "Testing by running the example" is working for you? Thank you very much!

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
```diff
@@ -96,6 +97,12 @@ namespace SoLoud
 
             case ma_device_notification_type_interruption_ended:
             {
+#if defined(MA_HAS_COREAUDIO)
+                // On macOS and iOS when the the interruption begins
+                // the device is automatically stopped (not uninited with ma_device_uninit).
+                // So we need to start it again when the interruption ends.
+                miniaudio_ensureDeviceStarted_impl()
```
MinseokKang003, line 104: @alnitak  Sorry, missed a trailing comma here. Please edit this.

MinseokKang003:
https://github.com/user-attachments/assets/464e0794-e5d9-4bc5-a9bc-d0c487f4cc64

https://github.com/user-attachments/assets/e3e512b8-941a-47a0-9920-feb753048d59

https://github.com/user-attachments/assets/e2aeec6b-4ffa-4c7f-8cd2-b3f2d11ac2c7

To test audio interruption in iOS, we need to set `AVAudioSessionCategoryOptions.none` for audio session configuration like below. If we use `AVAudioSessionCategoryOptions.mixWithOthers` we won't receive interruption notifications and the audio will always mix with other app's audio.
```
avAudioSessionCategory: AVAudioSessionCategory.playback,
avAudioSessionCategoryOptions: AVAudioSessionCategoryOptions.none,
```

Everything seems to work fine here.
Please let me know if you need more tests for other specific cases.

alnitak:
@MinseokKang003 this is great!! I am very grateful, thank you!

I think that, for now, we are ready to go. All the audio context seems to be managed only by `audio_session` now. Hoping the best :crossed_fingers:

MinseokKang003:
Yup no problems so far on my side. Feel free to merge! It's great to contribute.

themailman05:
+1 nice job

themailman05:
> Yup no problems so far on my side. Feel free to merge! It's great to contribute.

were you able to get this to work on iOS side? I'm not able to get playback still when implementing `audio_session` for session management.

themailman05:
I just tried to `flutter clean` and rebuild just in case and it still isn't working. @MinseokKang003 were you able to demonstrate recording + playback using `flutter_recorder` on iOS?

adventureisyou:
Hi. Sorry for the silence on my part, it's been a busy week! Thanks for all your work on this.

I've just tried using this branch for our app, and it seems to accomplish two things:
1. Avoid setting the audio category on launch.
2. Avoid starting/stopping audio sessions.

The first solves my issue of spotify being interrupted on startup. I'm now able to set my own audio category to ambient and it doesn't get overridden by miniaudio. This means YouTube/Spotify don't get interrupted when opening the app.

**INCORRECT SEE NEXT COMMENT**
~The second causes a bit of a headache. It works, and littering my code with all the manual `_session.setActive(true)` and `_session.interruptionEventStream.listen` calls necessary works, but feels like a lot of unnecessary work when it worked before. `audio_session` actually says that it expects the plugins to take care of starting the audio session when playing anything:~

>Each time you invoke an audio plugin to play audio, that plugin will activate your app's shared audio session to inform the operating system that your app is now actively playing audio.
...
>You normally do not need to activate the audio session yourself, however if the audio plugin you use does not activate the audio session, you can activate it yourself...

~I think this should only be merged if an option is exposed to control the start/stop session behaviour (or you do a major version bump with a breaking change notice). Separating out the first change and getting it merged would be my preference because it's a much less aggressive change. Users not managing audio category themselves will change from mediaPlayback to soloAmbient (the default), which isn't a huge change. Alternatively, this could be exposed as an option too.~

Finally, I think I need to apologise for conflating my startup focus issue with the original reports of audio session interruptions causing problems. My experiece is that even without this change, I'm unable to repro any of the symptoms in the original ticket. Both `Soloud.play()` and `Soloud.setPause(false)` cause the audio session to be grabbed again after an interruption. Sorry!

~Of course, this PR gives people the freedom to manage the audio session themselves, but like I say above it feels like a tool most users don't need and thus should probably be behind a setting.~

adventureisyou:
Gahh - maybe I'm talking rubbish. I tried this branch again without manual audio session management and it seems to work fine - though I don't understand how.

adventureisyou:
OH - the `noAudioSessionActivate` only affects initialisation. I guess play() and setPause() still do their thing of grabbing the session. In which case, everything works!! Sorry for sowing confusion. So many combinations to test I think I just got myself confused.

alnitak:
Your comment appeared just 2 seconds before I wanted to hit the "Merge" button :smile: !

Anyway, no problem at all, you are welcome to tell me your point of view and suggestions!!

Agreed with the fact that there are a lot of combinations to test and cannot be automated (I think), and I can't tell how many times I got confused too :)

If you need some more time to test your use cases, please tell me. I have no problem waiting to merge. I'd prefer to be sure that all is working fine. If you are just fine with this PR, I can merge and publish on pub.dev now. Take all the time!

adventureisyou:
Hah, sorry. I was writing it while testing and then saw you merge master back in and thought I'd post it quickly to stop you merging. Then realised I was wrong!

I'm very happy, let's merge! Thanks again.

PR #260, opened 2025-06-26
Ensure device started

MinseokKang003:
## Description
This PR is a continuation of the work from PR #257. 

This PR introduces a new `miniaudio_ensureDeviceStarted()` function. The function checks the device state and starts it if is stopped by interruptions, for `play` and `setPause` operations.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


alnitak:
Thank you very much for your contribution! Much appreciated!

I just changed the destination branch from `alnitak:main` to `alnitak:audioContext`.

Your intuition looks good to me. I was thinking the same, but with the check done in `Player::isInited()` method in `player.cpp` and not only in `setPause` and `play`. Because maybe restarting the device could be needed also when calling some other methods (like `play3d()`).

If you see in `bindings.cpp`, in almost all the functions, the `player.get()->isInited()` is called to check the state of the player. So, it could be a good place to put the check, but I am not sure because that method is called many times also by functions that don't need the check.

So, to recap, I would touch the SoLoud lib sources (the content in `src/soloud/src/`) as little as possible. Defining `miniaudio_ensureDeviceStarted()` is ok where you did it, but I'd move that check in those methods that need that in `player.cpp` (I think you will be able to call it with `soloud.miniaudio_ensureDeviceStarted()`).

Do not hesitate to give me suggestions or your point of view!!

MinseokKang003:
Yeah please just consider this PR as a reference. It's just a draft written with Claude. It's better for you to implement this since I don't have much understanding to C++ and the repo itself. Just let me know if you need more tests!

alnitak:
> Yeah please just consider this PR as a reference. It's just a draft written with Claude. It's better for you to implement this since I don't have much understanding to C++ and the repo itself. Just let me know if you need more tests!

But I'll be happy if your contribution could appear in the repo's Contributors list! ;)

I could try to do this in a couple of hours and see if it compiles and runs fine on Android. Then I'll commit and you will be able to test on iOS. I really hope to have this working!! :crossed_fingers:

MinseokKang003:
Haha okay let me try then. I've moved miniaudio_ensureDeviceStarted logic to player.cpp.
For now, I've added the function to `setPause`(only for False), `play`, `play3d`, `textToSpeech` functions individually.
Let me know if you think it's better to put it into `isInited`. I'll put it there then.

alnitak:
Ok thank you! Looks good to me.

I'll merge this as is. I think it is enough for now having the check only where you put them now. I'll try to put it in `isInited` and see its behavior on Android (it should be the same as on iOS).

I'll update you.

Issue #312, opened 2025-09-17
fix: stop() takes too much time


filiph:
**Description**

I keep seeing a log with “`stop() takes too much time for handle`" when running my flutter_soloud-using game.

I wanted to provide a failing test but I've been busy so I'll at least file this bug.

I've tracked the problem to something like this. I get the problem when:

- playing a lot of “single channel” sounds - new sounds replace preceding sounds
  - I'm doing this programatically (NOT using Soloud's voice feature). This is because I want to fade out the preceding sound while the new sound is playing.
- I don’t ignore the future returned by \`SoLoud.stop(handle)\`
- I spam starting the new sounds

I’m pretty sure there’s a bug. Some sound handles get through the cracks and their `voiceEndedCompleters[handle]?.complete()` never gets called.

**Steps To Reproduce**

I wish I had a better repro, but here's how I can do it in the `giant_robot` project (which @alnitak has access to):

- open the game
- wait until SFX are loaded (this takes a while in debug mode)
- start moving the mouse pointer up and down the menu, so that it produces a bunch of "bleep" sounds in quick succession

**Expected Behavior**

All `voiceEndedCompleters` are completed.

alnitak:
Hi Filip,

the problem occurs in some rare conditions when a `stop` is called while another `stop` hasn't triggered yet the voice ended callback from C to Dart. It's a race condition.

Looking at `giant_robot` code, I see in `soloud_audio_controller.dart` line 817 that the `stop` is awaited, but in `audio_controller.dart` line 434, it doesn't await. Here comes the rare race condition. But even if you await the stop, the problem persists. By commenting out one of those, the problem disappears.

This probably could be fixed on the C side, because by calling the voiceEndedCallback in the `stop` function, instead of waiting for the engine to do that, the problem goes out.

I tried hard to reproduce this without success!!

FYI, the button sound never fades out. When debugging, I never entered line 819 of `soloud_audio_controller.dart`.

filiph:
Thanks Marco for looking at this!

I wonder, is it possible that `_isHandlePresent()` on `soloud.dart:371` returns `null` despite `voiceEndedCompleters` having that handle? In other ways, is it maybe possible for `_activeSounds` and `voiceEndedCompleters` to get out of sync?

Because if so, then `voiceEndedCompleters[SoundHandle(handle)]?.complete();` on `soloud.dart:390` never gets called for that handle. I'd start with asserting that:

```dart
[...]

PR #317, opened 2025-09-18
Improve somewhat on the test, plus a stab at working better with futures.

filiph:
Concerns #312


alnitak:
After the #319, I'm asking if you think the mods you have done with `soloud.dar` are still valid.

I'd leave your changes and merge it.

Regarding the test, I have [modified](https://github.com/alnitak/flutter_soloud/pull/319/files#diff-678e50ef9a5f82cba22f58cbd52bfc0e15a4a9e54c87bd9fce7cc9d17b4cf1c5) it using just 1 play/stop and a `scheduleStop`.

If this sounds good for you, please revert `test.dart` and I'll merge this PR.

filiph:
Yeah, I think we can close this. There's some more logging here and possibly a solved edge case with closing a completer, but I'd rather work on #319 and see if it solves issues, and maybe resurrect this change as a nice-to-have followup.

PR #319, opened 2025-09-19
fix: stop() takes too much time

alnitak:
## Description

As discussed in #312, calling `stop` immediately after calling `play`, produces the warning.

2 problems were causing the issue:
1. the one discussed in the above issue: the `play()` doesn't have the time to produce a handle before the `stop()` is called.
2. a `stop()` call after another event like `scheduleStop` or a sound ended that already stopped the handle, ie:
```
final handle = await SoLoud.instance.play(currentSound);
SoLoud.instance.scheduleStop(handle, const Duration(seconds: 3));
await delay(3500); // wait till the handle has already been stopped by scheduleStop
unawaited(SoLoud.instance.stop(handle));
```

This fixes both problems.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/bindings.cpp
+++ b/src/bindings.cpp
@@ -115,8 +115,16 @@ extern "C"
             if (isHandleFound)
                 player->removeHandle(*handle);
             else
+                // If the handle is not found, for sure it is already
+                // removed by a previous call to `voiceEndedCallback`.
+                // For example triggering a `stop` in a `Future` after the sound is ended or
+                // vice versa.
                 return;
         }
+
+        // Here the internal flutter_soloud handle, doesn't exist anymore, whether this
+        // callback is called directly from `stop`, or whether it is called from an
+        // event in `Soloud::stopVoice_internal (unsigned int aVoice)`.
 
 #ifdef __EMSCRIPTEN__
         // Calling JavaScript from C/C++
@@ -128,9 +136,12 @@ extern "C"
         // The `dartVoiceEndedCallback` is not set on Web.
         if (dartVoiceEndedCallback == nullptr)
             return;
+        // So, if the handle was already found before (henche the handle is not found), the
+        // callback to Dart has been already called. If this is the fist time this handle 
+        // is found, the callback to Dart must be called.
         if (!isHandleFound)
             return;
-        // [n] pointer must be deleted on Dart.
+        // [n] pointer must be deleted in Dart.
         unsigned int *n = (unsigned int *)malloc(sizeof(unsigned int *));
         *n = *handle;
         dartVoiceEndedCallback(n);


src/bindings.cpp
```diff
@@ -108,8 +108,15 @@ extern "C"
...
     {
+        bool isHandleFound;
         if (player != nullptr)
-            player->removeHandle(*handle);
+        {
+            isHandleFound = player->findByHandle(*handle) != nullptr;
+            if (isHandleFound)
+                player->removeHandle(*handle);
+            else
+                return;
```
filiph, line 118: This is getting a bit too complex for me and I'd love to have a short comment here. Something like:

```suggestion
                // When player exists but handle isn't found,
                // bail out immediately.
                return;
```

So, when player _isn't_ null and we call this callback, we continue executing this function?

filiph: As a more general point, maybe elaborate on the logic here, because it's a bit too dense for my brain. The fact that there's a return statement in one of the branches makes it hard to reason about, imho. But maybe it's just a matter of elaborating on the comment suggested above?

src/bindings.cpp
```diff
@@ -121,6 +128,8 @@ extern "C"
         // The `dartVoiceEndedCallback` is not set on Web.
         if (dartVoiceEndedCallback == nullptr)
             return;
+        if (!isHandleFound)
```
filiph, line 131: If `player` is null, then `isHandleFound` is undefined (i.e. could be anything) at this point, right? Or does C++ set bools to false / `0` at init?

src/bindings.cpp
```diff
@@ -770,10 +779,10 @@ extern "C"
     /// [handle]
     FFI_PLUGIN_EXPORT void stop(unsigned int handle)
     {
-        if (player.get() == nullptr || !player.get()->isInited() ||
-            !player.get()->isValidHandle(handle))
```
filiph, line 774: Wait, why are we no longer checking whether the handle is valid? Does `Player::stop()` not care?

Oh wait, we checked whether the handle was _not_ valid? And then we stopped it? Sorry, I'm getting confused.

src/player.cpp
```diff
@@ -634,7 +634,6 @@ PlayerErrors Player::play(
 
 void Player::stop(unsigned int handle)
 {
-    removeHandle(handle);
```
filiph, line 637: [nit] I guess I'd like to see it documented (in comments in this file, ideally) who is responsible for removing the handle (e.g. calling `Player::removeHandle()`).

filiph (commented):
Hey Marco, thanks for the quick turnaround! I have a few questions below.

alnitak:
Hey Filip, sorry for the late.

I agree, I'm stingy with comments. I have added some in the callback. If some more are needed, please let me know.

src/bindings.cpp
filiph: Thanks for the explanation!

filiph (approved):
LGTM, thanks for adding the additional comments!

PR #344, opened 2025-10-24
Fix Opus BufferStream end clicks

### LINKED ISSUE #343: fix: Click sound at the end of OGG (Opus/Vorbis) files
When I play an OGG file that has been loaded in with the buffer stream as described in #306, I hear a small "click" sound at the end of the file playback. The click sound seems to happen randomly (it doesn't happen 100% of the time, and each time it sounds a bit different).

Do we have any clue what the root cause of the issue is? @alnitak 

Would be happy to also do my own investigation of the source code to figure it out!

Relevant comment:

> With the new 3.3.4 version, this has been fixed, but I will look further for the clicks at the end of small Opus or Vorbis files. 

 _Originally posted by @alnitak in [#306](https://github.com/alnitak/flutter_soloud/issues/306#issuecomment-3276073343)_

eddyleelin:
## Description

Fix "click" sound at the end of OGG Opus files (#343):
- fix the BufferStream Opus path so it honors the stream’s `pre_skip`, granule position, and declared sample rate/channels
- root cause: the previous decoder always ran at the engine rate (often 24 kHz), never applied `pre_skip`, and trimmed the PCM buffer to the raw granule position, which cut off real samples and produced the audible click
- solution: recreate the decoder at the stream’s rate/channels, drop the `pre_skip` frames before emitting PCM, and stop only after the expected sample count; the existing 5 ms fade now smooths the true tail instead of hiding a truncation
- confirmed this runs on top of the 3.3.5/3.3.7 audio-quality regression fix (no additional EQ or “dampened” artifacts show up)

✋ WARNING: the changes and commit were generated with the Codex CLI; I’m not an audio-domain expert, so please give this a thorough review! I don't know if all of the changes in the commit are absolutely necessary, but when I tried to simplify the changes or patch less (again, with little personal expertise and the help of AI), the clicks came back. Would be happy if this gets either merged or closed and just used as inspiration for a better fix! 😄 

Also, this doesn't touch the Vorbis stream decoder, which might need its own patch?

## Type of Change

## Testing

- streamed multiple short Opus clips through `setBufferStream` + `addAudioDataStream` in my own Flutter app using these changes patched locally to `flutter_soloud`. The clicks at the end of the audio are gone.

Fixes #343

### DIFF
--- a/src/audiobuffer/opus_stream_decoder.cpp
+++ b/src/audiobuffer/opus_stream_decoder.cpp
@@ -1,22 +1,78 @@
 #if !defined(NO_OPUS_OGG_LIBS)
 
 #include "opus_stream_decoder.h"
+#include <algorithm>
 
 OpusDecoderWrapper::OpusDecoderWrapper()
-    : streamInitialized(false), headerParsed(false), packetCount(0)
+    : decoder(nullptr),
+      engineSamplerate(48000),
+      engineChannels(2),
+      decodingSamplerate(48000),
+      decodingChannels(2),
+      streamInitialized(false),
+      headerParsed(false),
+      packetCount(0),
+      skipSamplesPending(0),
+      totalOutputSamples(0),
+      totalSamplesExpected(-1)
 {
 }
 
 OpusDecoderWrapper::~OpusDecoderWrapper()
 {
     if (decoder)
+    {
         opus_decoder_destroy(decoder);
+        decoder = nullptr;
+    }
 
     if (streamInitialized)
     {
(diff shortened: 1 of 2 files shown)


alnitak:
I have tested this and the code seems good to me!

> Also, this doesn't touch the Vorbis stream decoder, which might need its own patch?

This is fixing the Opus stream. If you are experiencing clicks also with Vorbis, please let me know in a new issue.

Thanks a lot for the work you have done! Merging.

PR #391, opened 2025-12-30
Fix MP3 streaming decoder audio loss with small chunk sizes

nukes:
## Description

 Problem

  The drmp3 high-level API (drmp3_read_pcm_frames_f32) has inherent issues with streaming:
  - When drmp3_init is called, it consumes all buffered data to find the first valid frame
  - Small chunks (< 2KB) are treated as EOF, causing the decoder to enter an unrecoverable state
  - Only chunks ≥ 2KB work reliably, but real-world streaming often delivers smaller chunks

  Test Results with High-Level API:
  | Chunk Size | Samples Decoded | Audio Loss |
  |------------|-----------------|------------|
  | 64 bytes   | 1,152           | 99.99%     |
  | 256 bytes  | 1,152           | 99.99%     |
  | 1024 bytes | 1,152           | 99.99%     |
  | 2048 bytes | 8,470,656       | 0%         |

  Solution

  Replace the drmp3 high-level API with the low-level drmp3dec (minimp3) API, which:
  - Decodes frame-by-frame without maintaining problematic internal state
  - Works correctly with ALL chunk sizes (64B to 64KB tested)
  - Has minimal state (only 4 bytes) making it ideal for streaming
  - Properly handles partial frame data by accumulating in an input buffer

  Test Results with Low-Level API:
  | Chunk Size  | Samples Decoded | Audio Loss |
  |-------------|-----------------|------------|
  | 64 bytes    | 8,469,504       | 0%         |
  | 256 bytes   | 8,469,504       | 0%         |
  | 1024 bytes  | 8,469,504       | 0%         |
  | 65536 bytes | 8,469,504       | 0%         |

  Changes

  - mp3_stream_decoder.h/cpp: Replaced drmp3 with drmp3dec low-level decoder
  - audiobuffer.cpp: Added mutex protection for buffer access
  - setDataEnded(): Now a no-op for MP3 (low-level API doesn't need it)
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/audiobuffer/audiobuffer.cpp
+++ b/src/audiobuffer/audiobuffer.cpp
@@ -492,7 +492,10 @@ namespace SoLoud
 			}
 			else
 				// This handle has reached [TIME_FOR_BUFFERING]. Unpause it.
-				if (currBufferTime + addedDataTime - mParent->handle[i].bufferingTime >= mBufferingTimeNeeds && isPaused)
+				// Only unpause when buffer covers playback position + margin,
+				// not just when new data >= margin (which caused play/pause toggling
+				// when seeking beyond buffered data)
+				if (currBufferTime + addedDataTime >= pos + mBufferingTimeNeeds && isPaused)
 				{
 					mThePlayer->setPause(handle, false);
 					isPaused = false;


alnitak:
Hi @nukes, yesterday I've updated the plugin with PR #390 that was maybe related to this one. Is the problem you noticed present also after that merge (the version 3.4.8 on pub.dev)? I think so, but it’s hard for me to tell what changed since the indentation makes every line appear modified.

The minimum chunk size that `MP3DecoderWrapper::decode` is 32 KB (which is delivered from [here](https://github.com/alnitak/flutter_soloud/blob/9a6250701485e644b69bd8291ce35097208fb8e1/src/audiobuffer/audiobuffer.cpp#L316)), except the last few KB, which are delivered when calling `setDataIsEnded`. I tried using [websocketd](https://github.com/alnitak/websocketd) and the [`websocket.dart`](https://github.com/alnitak/flutter_soloud/blob/main/example/lib/buffer_stream/websocket.dart) example using chunks from 32 bytes to 2048. I didn't noticed any audio loss (just hearing the audio). So I didn't understand well where the problem is. Can you please give me some more info and maybe an audio sample to test and maybe a bit of code?

I suspect you have used some AI because for example there is no more `MP3DecoderWrapper::on_meta` method and it is mandatory here when initializing dr_mp3 to get MP3 metadata when available.

nukes:
Hi @alnitak   I originally want to fix the data loss issue when do mp3 streaming decoding. i just tested you new code from main branch and i found you have already solved it. let's close the PR.
I can still find a seek issue when streaming. i will start a new PR.
think you!

PR #396, opened 2026-01-21
Elevate audio thread priority on Android

djkingCanada:
### Problem

On many Android devices, the audio thread spawned by MiniAudio is categorized as a background or batch process (SCHED_BATCH). This leads to significant scheduling delays (50ms+), causing audible static and "stuttering" even when CPU usage is low. This is especially prevalent when the app is under load or when targeting older API levels.

### Solution
This PR implements a manual priority elevation within the MiniAudio backend. On the first invocation of the audio callback (which runs on the dedicated audio thread):

Escapes SCHED_BATCH: If the thread is found to be in the Batch policy, it is moved to SCHED_NORMAL (Policy 0).
Elevates Priority: Attempts to set the thread to Realtime (SCHED_FIFO) or Round-Robin (SCHED_RR).
Niceness Adjustment: Falls back to setting the highest possible niceness (-20) using setpriority if Realtime access is denied by the OS.
These changes are wrapped in #ifdef __ANDROID__ guards to ensure zero impact on other platforms.

### Impact
Stability: Prevents the OS from preempting the audio thread.
Latency: Reduces "jitter" and audio underruns.
Quality: Eliminates persistent "static" noise on problematic Android hardware.

### Testing
Verified on Android devices where previous versions experienced consistent static. Jitter monitoring showed the thread meeting its deadlines (e.g., 2048 samples @ 48kHz) consistently after the fix.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
+++ b/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
@@ -58,6 +58,9 @@ namespace SoLoud
 // #define MA_DEBUG_OUTPUT
 #include "miniaudio.h"
 #include <math.h>
+#include <chrono>
+#include <pthread.h>
+#include <sys/resource.h>
 
 namespace SoLoud
 {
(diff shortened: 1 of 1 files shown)


alnitak:
Thank you @djkingCanada! This is a great improvement!

I would merge this into the `dev` branch for a while. I have already tested on a new and on an old device, and it works great, even if I didn't notice stuttering before (but maybe I don't have the right device for that).

I will test again in the next few days before merging into `main`.

Thanks again.

alnitak:
I have published `v3.5.0-pre.0` on pub.dev with both this and #397 fixes

PR #397, opened 2026-01-21
Enable AAudio with runtime API level check for safe fallback

djkingCanada:
### Problem
Currently, soloud_miniaudio.cpp contains a compile-time guard (MA_NO_AAUDIO) that disables the AAudio backend if __ANDROID_API__ is 29 or lower.

This was originally added (Commit d59c5b4) as a workaround for a specific crash during  deinit() reported on some Android 10 devices.

However, because __ANDROID_API__ is typically set to the project's minSdkVersion at compile-time, this guard effectively disables AAudio on every device the app runs on—including modern Pixels running Android 14—if the project supports any device older than Android 11. This forces a legacy OpenSL ES fallback across the entire user base, regardless of each individual device's actual capabilities.

### Solution
This PR replaces the global compile-time ban with an intelligent runtime check:

**Remove Compile-time Guard**: Deletes the MA_NO_AAUDIO restriction based on __ANDROID_API__.
**Runtime API Check**: Uses android_get_device_api_level() during engine initialization.

Inteligent Fallback:
If the running device is API 29 (Android 10) or lower, we explicitly tell MiniAudio to use OpenSL ES. This preserves the original workaround for the reported  deinit() crash on those specific versions.

If the device is API 30 (Android 11) or higher, we allow MiniAudio to use AAudio as the preferred backend.
Impact

Performance: Significant reduction in audio latency on Android 8-10.
Stability: Better integration with modern Android audio features like performance profiles.

**Testing**
Tested on devices targeting API 21 and 29. Verified via logging that the system correctly negotiates the best available backend (AAudio where supported, OpenSL otherwise) and remains stable during intense playback.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
+++ b/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
@@ -52,11 +52,14 @@ namespace SoLoud
 // addressed by this issue: https://github.com/mackron/miniaudio/issues/466
 // For me this happens using AAudio on android <= 10 (but not on Samsung Galaxy S9+).
 // Disablig AAudio in favor of OpenSL is a workaround to prevent the crash.
-#if defined(__ANDROID__) && (__ANDROID_API__ <= 29)
-#define MA_NO_AAUDIO
-#endif
+// #if defined(__ANDROID__) && (__ANDROID_API__ <= 29)
+// #define MA_NO_AAUDIO
+// #endif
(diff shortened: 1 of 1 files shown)


alnitak:
Thank you very much @djkingCanada. I much appreciate this fix!

alnitak:
I have published v3.4.9 on pub.dev with this fix

Issue #375, opened 2025-11-26
fix: PathNotFoundException init()


Solido:
**Description**

Hi!

Under macos desktop app:

After a version update, init() throw an PathNotFoundException related to:
<UserPath>/Data/Library/Caches/com.board.luna/SoLoudLoader-Temp-Files

**tmp** is present but not **SoLoudLoader-Temp-Files** and it's not created, audios are not copied.

This 'seems' to be related to a lib update to last version.

Thank you!

alnitak:
Hi @Solido,

as a premise, I should say that I am not very familiar with MacOS.

I see the `SoLoudLoader-Temp-Files` in
```
Library/Containers/<your_app_name>/Data/Library/Caches
```

so I think you should find it in
```
Library/Containers/com.board.luna/Data/Library/Caches
```

Let me know!

Solido:
Yep that's the dir. I had to recreate it manually to get back to work.
Curious case but would like to share so maybe you can put a guard.
Do you have the right to check and create the dir if not existent? 
Else it means a permanent corrupted installation.

Thanks!

alnitak:
Could it be a `path_provider` problem? We use that to retrieve that folder and create `SoLoudLoader-Temp-Files`. The `Data/Library/Caches` should be an app sandboxed folder with read/write permission!

alnitak:
> Yep that's the dir. I had to recreate it manually to get back to work.

ah, I had explained it badly. You wrote that the folder is not in `/Data/Library/Caches/com.board.luna/SoLoudLoader-Temp-Files` but it should instead be found in `/Data/Library/Caches/com.board.luna/Data/Library/Caches/SoLoudLoader-Temp-Files`

Solido:
Confusing! here's the missing path reported by the plugin:

`/Users/username/Library/Containers/com.board.luna/Data/Library/Caches/com.board.luna/SoLoudLoader-Temp-Files`

After mkdir, Solound is now working again. 

I don't get the second path, why is there 2 Caches dirs?

alnitak:
Because I'm freaking out 😃 

Your dir is almost the right one (I don't think the 2nd `com.board.luna` is correct), but I don't explain why you have to create the dir manually!
```
/Users/<user>/Library/Containers/<your_app_ID>/Data/Library/Caches/SoLoudLoader-Temp-Files
```

Solido:
I close the ticket as we have a clear idea of what's going on... except why the suppression.
[...]

PR #404, opened 2026-02-25
Harden loader temp directory logic

filiph:
## Description

This change addresses a few issues with the current I/O loader:

- It no longer tries to delete the temprorary directory in `cleanUp`. This would lead to the directory being deleted just after initialization when `automaticCleanup` is `true`, which is a catastrophic failure.
- It creates the temporary directory with `recursive: true`, trying to avoid failures when the directory returned by `path_provider` doesn’t yet exist.
- If the directory supplied by `path_provider` fails for any reason (I'm seeing this in the wild on macOS), use the system temp directory instead.
- It adds stack trace to the severe log in initialization, and downgrades cleanup severe logs to warning logs.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/lib/src/utils/loader_io.dart
+++ b/lib/src/utils/loader_io.dart
@@ -70,11 +70,12 @@ class SoLoudLoader {
       // Delete files in parallel.
       final results = await Future.wait(files.map(deleteFile));
 
-      if (results.any((success) => !success)) {
-        _log.warning(
-          'Cannot clean up temporary directory. See warnings above.',
-        );
+      final successes = results.where((success) => success == true).length;
+      if (successes < files.length) {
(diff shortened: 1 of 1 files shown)


filiph:
No hurry to land this. I'll test it in the wild first, to see if it actually makes things better.

alnitak:
That's great @filiph! This comes in time for the (already closed!) issue in [this post](https://github.com/alnitak/flutter_soloud/issues/375#issuecomment-3955391743).

Let me know when you feel like merging.

filiph:
I think this is ok — I haven't seen any reports from the wild.

alnitak:
Great! Thank you!

PR #405, opened 2026-02-25
feat: Mixing buses

### LINKED ISSUE #364: feat: Sidechain ducking
**Description**

When playing sound effects and music at the same time, sidechain ducking is a useful feature.

Basically, here's how this feature works. Let's take a video game with background music and SFX as an example:

1. We put all SFX into a single audio bus ("Mixing Bus" feature of Soloud, see here: https://solhsa.com/soloud/concepts.html)
2. We monitor the volume on that bus
3. We use the _SFX volume_ as the input of a compressor on the _music_ volume. In other words, when the volume of all the sound effects (combined) crosses some threshold, we lower the _music volume_. The music volume thus "ducks" down whenever the sfx volume is loud.

This lets the music be loud when the SFX is quiet while also making sure that sound effects can be heard without distortion when they are loud.

Sidechain / ducking is also popular for mixing voice. The voice provides the sidechain, and music or sound effects "duck". Think of a radio DJ lowering the volume of a track to talk over it, except this is automatic and usually much faster.

**Requirements**

- [ ] Voices can be put into Mixing Buses
- [ ] The compressor filter (or its derivative) can perform its work based on audio output of a mixing bus (through a side chain) instead of the output of its own channel
- [ ] Make sure this is performant enough (we probably don't want this feature if it means some kind of drastic slow down)
- [ ] Some kind of test that this works

**Additional Context**

- FMOD: https://www.fmod.com/docs/2.02/studio/effect-reference.html#compressor
- Ableton live: https://www.ableton.com/en/blog/elphnt-introduction-sidechaining-live/ has a nice video that explains sidechaining quite well

alnitak:
## Description

This PR branch relies on the `dev` branch, which is almost ready to be released (see the `pre` versions in CHANGELOG.md).

https://solhsa.com/soloud/mixbus.html
The mixing buses are a special case of an audio stream. They are a kind of audio stream that plays other audio streams. Mixing bus can also play other mixing buses. Like any other audio stream, mixing bus has volume, panning and filters.

Only one instance of a mixing bus can play at the same time; however, trying to play the same bus several times stops the earlier instance.

2 new classes have been added:

* **`Bus`**: Represents the mixing bus itself. It manages its own ID, filters, and playback state (like `soundHandle`). You can execute bus-specific operations like setting channels, getting the active voice count, or moving an already playing sound to it (`annexSound()`).
* **`Buses`**: A lightweight singleton helper class used to track all active `Bus` instances globally. It provides handy lookup methods like `byName()` and `byId()`.

**Workflow**

1. **Create the Bus**
   Create a new bus instance via the SoLoud engine.
   `final myBus = SoLoud.instance.createMixingBus();`

2. **Play it on the Engine**
   Before you can hear any sounds routed through the bus, you must make the bus audible on the main engine.
   `myBus.playOnEngine();`

3. **Route Sounds Through It**
   Instead of default playback on the main engine, use the `play()` or `play3d()` function on the `myBus` instance itself!
   `final sound = await SoLoud.instance.loadAsset('audio.mp3');`
   `myBus.play(sound);`
   To the usual SoLoud `play()` and `play3d()`, has been added the `busId` parameter which informs the engine to play that sound on a specific bus. The counterpart methods in the Bus class are just a comfy way to call the SoLoud methods from within the Bus class.

Now, applying filters (`myBus.filters.echoFilter.activate()`) or adjusting the handle's volume will seamlessly apply to every audio source currently routed through this mixing bus.

**Considerations**

The `playOnEngine` method returns the `SoundHandle` of the mixing bus. For convenience, this handle is also stored inside the `Bus` class, and can be used with all the `SoLoud` methods which take it as a parameter (`stop`, `seek`, `setPause`, `set3dSourceParameters`, etc.).

By creating 2 different buses and using the same `AudioSource` in both buses, will produce some inconveniences:
- the `Bus.getActiveVoiceCount` did not return the correct number of voices
- if it is needed to listen for sound events (`AudioSource.soundEvents.listen` or `AudioSource.allInstancesFinished.listen`) we don't know the bus where the event comes from

Think if a class is used to manage a bus to listen for enemy sounds (walk, shots, etc.). There could be many enemies, each managed by different buses. This class will load the sounds for each instance (enemy). All these sounds will then have the same sound hash and thus be treated as the same for all instances.

To have a clearer idea, look at the `mixing_bus/mixing_bus.dart` example.
To solve this, it is now possible to load the same audio file more than once. The warning in the console of duplicate sound loaded is still displayed, but it is now responsability of the dev to choose what to do.

The `mixing_bus/mixing_bus.dart` example shows how to create multiple buses, play sounds through them, set volume and a filter independently, attach a sound that is playing in the main player to a bus.

- [x] Web support

/cc @filip, @NTayloe

fixes #364 

## Type of Change


NTayloe:
This is awesome, thank you! I'm excited to start messing around with buses, they should help fix a lot of minor issues in my project.

alnitak:
This should be good to merge, but if you have any suggestions (especially around naming or structure), feel free to drop them — I’m happy to tweak things before it goes in 🙂

PR #406, opened 2026-03-07
fix: wire miniaudio backend pause/resume to stop AudioUnit on iOS

sbauly:
## Problem

When using `flutter_soloud` alongside `audio_service` on iOS, the lock screen controls and Bluetooth headset controls (e.g. AirPods) become permanently unresponsive after pausing audio - unless the pause was initiated by the lock screen itself.

The only way to resume playback was via the app UI directly. The controls remain visible and non-greyed-out, but pressing play does nothing and no events fire. I have attached videos of this bug in action.

**This does not happen with `just_audio`** under the identical `audio_service` implementation, which points directly to SoLoud's pause behaviour as the cause.

## Root Cause

`SoLoud.setPause()` is a pure mixer flag - it stops audio samples from being mixed, but the underlying CoreAudio `AudioUnit` keeps running and outputting silence.

iOS interprets a running `AudioUnit` that produces silence as inactivity. With Bluetooth audio in particular, silence-detection causes iOS to demote the app from "now playing" status and stop routing `MPRemoteCommandCenter` events to it. `AVPlayer.pause()` avoids this because Apple's media framework explicitly signals the OS that the app is paused; raw CoreAudio does not.

The deeper issue: the miniaudio backend never assigned `mBackendPauseFunc` / `mBackendResumeFunc`. Those hooks exist in `soloud.h` and the CoreAudio backend uses them, but miniaudio left them `NULL`, meaning `Soloud::pause()` was silently returning `NOT_IMPLEMENTED` at the backend level.

## Fix

1. **`soloud_miniaudio.cpp`** - adds `soloud_miniaudio_pause()` and `soloud_miniaudio_resume()`, wired into `mBackendPauseFunc` / `mBackendResumeFunc`. These call `ma_device_stop()` / `ma_device_start()` - stopping the CoreAudio AudioUnit cleanly without tearing down and reinitialising the device.
2. **`bindings.cpp`** - exposes `pauseAudioDevice()` / `resumeAudioDevice()` via FFI.
3. **`bindings_player.dart`, `bindings_player_ffi.dart`, `bindings_player_web.dart`** - Dart FFI bindings.
4. **`soloud.dart`** - public `SoLoud.pauseAudioDevice()` / `SoLoud.resumeAudioDevice()` API.

## Usage

Callers using `audio_service` (or managing pause/resume manually) need to pair the new calls 
with `setPause()`:

- Call `pauseAudioDevice()` immediately after `setPause(handle, true)`
- Call `resumeAudioDevice()` immediately before `setPause(handle, false)` or `play()`

`resumeAudioDevice()` is safe to call on first play before any pause has occurred.

> **Note:** The current implementation requires callers to manually pair `pauseAudioDevice()` / 
> `resumeAudioDevice()` with `setPause()`. It may be cleaner to internalize this pairing inside 
> `setPause()` itself - happy to explore that if preferred.

## Evidence

I tested this using the `audio_service` official example app with `just_audio` and `flutter_soloud` made interchangeable via a flag (example code available on request). With the exact same audio session implementation, soloud was unable to resume audio playback after pausing via the app's UI or AirPods - but this worked fine with just_audio.

Of course you cannot see in these videos, but from the control center I was toggling play/pause by using AirPods built-in controls (not actually tapping the screen).

- 🎥 **Recording 1**: `just_audio` - AirPods pause/resume works correctly from control center.
https://github.com/user-attachments/assets/f6ddeee9-5e00-4421-b6c6-a4615da686cc

- 🎥 **Recording 2**: `flutter_soloud` without fix - AirPods pause leaves lock screen controls permanently unresponsive.
https://github.com/user-attachments/assets/0bd4b552-dd8d-4e26-897f-8c389758f6ed

- 🎥 **Recording 3**: `flutter_soloud` with this fix - AirPods pause/resume works correctly.
https://github.com/user-attachments/assets/95dba1b1-e439-496e-b99d-cbad99c53f63

## Notes

- I was heavily assisted by Claude in the diagnosis and fix of this issue...
- I don't have deep experience with the miniaudio/CoreAudio layer, so I'd welcome review of the C++ side in particular.
- @alnitak - I know you don't have an iPhone to test with, so I'm happy to be a testing resource if you want to iterate on this.

Related: #104, PR #257

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+#### 3.5.2 (18 Mar 2026)
+- fix: wire miniaudio backend pause/resume to stop AudioUnit on iOS #406. Thanks to @sbauly
+
+#### 3.5.1 (14 Mar 2026)
+- `getStreamTimeConsumed` return wrong time for s16le and s8 #419
+- win fix: hang on app exit #413
+- win fix: prevent Windows message pump goes irresponsive with plugins like `desktop_drop` and maybe others #401
+- fix: only unpause when buffer covers playback position #393. Thanks to @nukes
+- wasm fix: runtimeType error when voice ended #414
+
 #### 3.5.0 (1 Mar 2026)
 - Harden loader temp directory logic #404. Thanks to @filiph
 - updated audio_context example
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -3,7 +3,7 @@ description: >-
   A low-level audio plugin for Flutter,
   mainly meant for games and immersive apps.
   Based on the SoLoud (C++) audio engine.
-version: 3.5.0
+version: 3.5.2
 homepage: https://github.com/alnitak/flutter_soloud
 maintainer: Marco Bavagnoli (@lildeimos)
 platforms:


alnitak:
A ton of thanks, I'd be extremely happy to solve this age-old and boring problem!!

> I know you don't have an iPhone to test with...

I think this week I'll buy a refurbished one :). 
I would just like to ask you to please fix and test the example in `example/lib/audio_context/audio_context.dart`.

Is this fix working on iOS only, or on macOS as well?

There are just a few nits that need to be addressed, but I'll address them myself in this PR.

Thanks again for your time spent on this!

lib/src/soloud.dart
```diff
@@ -365,6 +365,43 @@ interface class SoLoud {
...
+
+  /// Resume the audio device after [pauseAudioDevice].
+  ///
+  /// On iOS the `AVAudioSession` must already be active before calling this
+  /// (call `AudioSession.instance.setActive(true)` first).
+  ///
+  /// Returns 0 on success.
+  int resumeAudioDevice() {
+    return _controller.soLoudFFI.resumeAudioDevice();
+  }
```
alnitak: Just a nit to return a bool instead of an int

alnitak:
Hey @sbauly.

My brand new "refurbished" iPhone has arrived :)

It took me a while, but some lines of your code got my eyes open: the SoLoud lib can handle the pause and resume custom callbacks!

I think this issue is fixed now without the mandatory use of `pauseAudioDevice` and `resumeAudioDevice` for the users. It is done in the Player class when the pause or play methods are called (and somewhere else). The key point is that we can forget to call them on iOS.

I have also updated the `example/audio_context/audio_context.dart` to also use `audio_service` for testing the player when the app is in the background.

Would be wonderful if you could find some spare time to test these changes on your side with both the provided example and your test case.

sbauly:
Hey @alnitak,

That’s great you managed to get hold of an iPhone! I’m sure it will be very helpful for testing any future issues :)

I found some time this evening to test your changes using the `audio_context` example and they worked like a charm! 

I tested with AirPods and background/play pause events and the issues I was previously seeing with SoLoud seem to be resolved. It’s also a much cleaner fix than using separate `pauseAudioDevice` & `resumeAudioDevice` methods.

Thanks so much for your work on this! Pinning down audio context issues can be very fiddly.

alnitak:
Thanks @sbauly!!

I am releasing v3.5.2

PR #434, opened 2026-03-30
fix: setBufferStream fails to decode small MP3 files under 32 KB

### LINKED ISSUE #433: setBufferStream fails to decode small MP3 files under 32 KB
When the total data fed to a buffer stream is below 32 KB, only about 0.17 seconds of audio is decoded regardless of actual file duration. The same bytes loaded via loadMem play correctly.

This commonly affects short MP3 files such as text to speech output, which are typically 10 to 30 KB per sentence.

## Reproduction

1. Obtain or generate a short MP3 file under 32 KB. I've included a test fixture in my fix branch: MPEG2 Layer III, 24 kHz, mono, 18 KB, 3 seconds ([test_mpeg2_24khz_mono.mp3](https://github.com/chaudharydeepanshu/flutter_soloud/blob/fix/buffer-stream-small-mp3/example/assets/audio/test_mpeg2_24khz_mono.mp3)).
2. Feed the complete bytes to setBufferStream, addAudioDataStream, then setDataIsEnded.
3. handleIsNoMoreValid fires after about 170 ms instead of the expected 3 seconds.
4. Load the same bytes with loadMem. Plays the full 3 seconds correctly.

```dart
// Only ~170 ms of audio plays
final source = SoLoud.instance.setBufferStream(
  format: BufferType.auto,
  bufferingTimeNeeds: 0.3,
  bufferingType: BufferingType.preserved,
);
final handle = await SoLoud.instance.play(source);
SoLoud.instance.addAudioDataStream(source, mp3Bytes); // 18 KB total
SoLoud.instance.setDataIsEnded(source);
// handleIsNoMoreValid fires at ~170 ms

// Full ~3 s plays correctly
final source2 = await SoLoud.instance.loadMem('test', mp3Bytes);
await SoLoud.instance.play(source2);
```

## Root cause

The 32 KB buffering threshold in BufferStream::addData() delays the first decode() call until setDataIsEnded() forces it. Because decode() was never called, StreamDecoder has not detected the format and the decoder wrapper (MP3DecoderWrapper) does not exist.

BufferStream::setDataIsEnded() calls streamDecoder->setDataEnded() before the final addData(). But StreamDecoder::setDataEnded() only forwards the signal when mWrapper is non null. It is not, so the signal is dropped.

When the final addData() runs decode(), the wrapper is created and drmp3_init reads all available data via on_read. The decode loop then checks m_read_pos >= audioData.size() && !mDataIsEnded. Since mDataIsEnded was never set, the loop breaks instead of draining dr_mp3's internal buffer. Only the frames decoded during drmp3_init (about 0.17 seconds) are returned.

Files above 32 KB work because the wrapper is created during normal addData() calls before setDataIsEnded().

## Environment

flutter_soloud 3.5.4, macOS (Apple Silicon). Tested with MPEG2 Layer III 24 kHz mono and MPEG1 Layer III 44.1 kHz stereo.

I have a fix ready with a test: https://github.com/chaudharydeepanshu/flutter_soloud/tree/fix/buffer-stream-small-mp3

chaudharydeepanshu:
Fixes #433

## Problem

setBufferStream fails to fully decode MP3 files smaller than 32 KB. Only about 0.17 seconds of audio is produced instead of the full duration. See #433 for reproduction steps and root cause analysis.

## Fix

Store a pending flag (mDataEndedPending) in StreamDecoder when setDataEnded() is called and the decoder wrapper does not exist yet. Forward the flag to the wrapper when it is created in decode().

## Changes

**stream_decoder.h**: Add mDataEndedPending field. setDataEnded() always stores the flag and forwards to the wrapper if it exists.

**stream_decoder.cpp**: After creating the wrapper in decode(), check and forward the pending flag.

**mp3_stream_decoder.h and mp3_stream_decoder.cpp**: Rename mDataIsEnded to mDataEnded for consistency with setDataEnded().

**example/assets/audio/test_mpeg2_24khz_mono.mp3**: 3 second test fixture, MPEG2 24 kHz mono, 18 KB.

**example/tests/tests.dart**: Test that verifies both small (under 32 KB) and large (over 32 KB) MP3 files decode fully via setBufferStream.

## Type of Change

- [x] Bug fix (non breaking change which fixes an issue)

### DIFF
--- a/example/tests/tests.dart
+++ b/example/tests/tests.dart
@@ -1016,8 +1016,7 @@ Future<StringBuffer> testBufferStreamSmallMp3() async {
       .buffer
       .asUint8List();
 
-  final smallRefSource =
-      await SoLoud.instance.loadMem('small_ref', smallMp3);
+  final smallRefSource = await SoLoud.instance.loadMem('small_ref', smallMp3);
   final smallRefDuration = SoLoud.instance.getLength(smallRefSource);
   await SoLoud.instance.disposeSource(smallRefSource);
 
@@ -1044,8 +1043,7 @@ Future<StringBuffer> testBufferStreamSmallMp3() async {
       .buffer
       .asUint8List();
 
-  final largeRefSource =
-      await SoLoud.instance.loadMem('large_ref', largeMp3);
+  final largeRefSource = await SoLoud.instance.loadMem('large_ref', largeMp3);
   final largeRefDuration = SoLoud.instance.getLength(largeRefSource);
   await SoLoud.instance.disposeSource(largeRefSource);
 


alnitak:
Thanks @chaudharydeepanshu!

This PR fixed the problem well, LGTM.

I'll also try to lower the buffer from 32KB to maybe 16. IIRC, there were some problems with OGG streams with a lower buffer, but things have changed since I first chose the "32KB" way.

PR #441, opened 2026-04-09
version 4.0.1 chores

alnitak:
## Description

- fixed get eq params other than bands values
- removed some warnings from native build
- warn mac users to install cmake if not already installed

## Type of Change

- [X] 🗑️ Chore


ekuleshov:
> warn mac users to install cmake if not already installed

@alnitak with this version I'm getting a hard build fail when trying to build prod app for iOS. cmake is installed with brew and is in regular system path. Is there something else needed to be updated?

alnitak:
Hey @ekuleshov, I tried to run "flutter build ios --release" with success. I just added a check in the `podspec` file:
```
if ! command -v cmake &> /dev/null; then
  echo "Error: CMake is not installed. Please install CMake to build flutter_soloud."
  echo "  - On macOS: brew install cmake"
  echo "  - Or visit: https://cmake.org/download/"
  exit 1
fi
```

Are you seeing those echoes?

ekuleshov:
The `flutter build ios`, but I'm actually using Product / Archive action from xcode to upload the release builds. It shows this text in the build failure. 

image

alnitak:
got it! Xcode uses a restricted PATH environment, so we should add the common locations where cmake might be installed in the `podspec` file.

Thanks for pointing this out. I have fixed this in the `dev` branch and if this is working for you also, I will release a new version instantly!

ekuleshov:
> Thanks for pointing this out. I have fixed this in the `dev` branch and if this is working for you also, I will release a new version instantly!

@alnitak the dev branch works. thank you for looking into this.

alnitak:
Thanks to you! Updating...

PR #444, opened 2026-04-16
fix(init): rebind Dart callbacks after hot restart

skylartaylor:
## Description

Fixes a hot-restart bug where `SoLoud.isInitialized` could report that the engine was ready even though the current Dart isolate had not rebound its native callbacks yet.

After a Flutter hot restart, the native SoLoud engine survives but the old Dart isolate's callback bindings are gone. If app code skipped `init()` because `isInitialized` returned `true`, the next native callback would hit a stale Dart pointer and crash.

### What changed

**Dart side** (first commit):
- Track whether native callbacks have been registered for the current Dart isolate (`_nativeCallbacksInitialized`)
- `isInitialized` now requires both native engine state AND current-isolate callback readiness
- `init()` detects when native is already alive (hot-restart case), clears stale callback registrations, tears down, and re-initializes with fresh callbacks

**C++ side** (second commit):
- Global callback pointers (`dartVoiceEndedCallback`, `dartFileLoadedCallback`, `dartStateChangedCallback`) and per-stream callback pointers (`mOnBufferingCallback`, `mOnMetadataCallback`) are now `std::atomic` — prevents TOCTOU races where the audio thread reads a pointer that another thread is nulling
- All callback call sites use snapshot-then-call: `auto cb = ptr.load(); if (cb) cb(args);`
- Fixed a memory leak in `stateChangedCallback` where `malloc` was called before the null check

**Test** (second commit):
- Added `HotRestartLifecycle` on-device integration test matching the project's `example/tests/tests/` pattern
- Exercises: fresh init, deinit, clean re-init, hot-restart recovery (init while native is still alive), post-recovery callback verification (load + play + stop), and rapid init/deinit cycling (5x)
- Deleted mock-based unit test that didn't fit the project's test paradigm

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/example/tests/tests/all_tests.dart
+++ b/example/tests/tests/all_tests.dart
@@ -3,6 +3,7 @@ import 'all_instances_finished.dart' as all_instances_finished;
 import 'async_multi_load.dart' as async_multi_load;
 import 'asynchronous_deinit.dart' as asynchronous_deinit;
 import 'auto_dispose.dart' as auto_dispose;
+import 'hot_restart_lifecycle.dart' as hot_restart_lifecycle;
 import 'buffer_stream_extended.dart' as buffer_stream_extended;
 import 'buffer_stream_small_mp3.dart' as buffer_stream_small_mp3;
 import 'compressor_filter.dart' as compressor_filter;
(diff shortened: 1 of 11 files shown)


alnitak:
Hi @skylartaylor, thank you very much for your time for this PR!

I've tested this on all the platforms, and it works well.

Probably this will fix #369

PR #445, opened 2026-04-16
fix(ffi): retain BufferStream callbacks until disposal

skylartaylor:
## Description

Fixes a Dart FFI callback lifetime bug in `setBufferStream()`.

The `NativeCallable` wrappers created for `onBuffering` and `onMetadata` callbacks were dropped immediately after `setBufferStream()` returned. With no Dart-side reference keeping them alive, the GC could free the trampoline while native code still held the function pointer — leading to a crash on the next buffering state change or metadata event.

### What changed

- **Callback retention**: Per-stream `NativeCallable` wrappers are now stored in a `Map<int, _BufferStreamNativeCallbacks>` keyed by sound hash, keeping them alive for the lifetime of the stream.
- **Cleanup on dispose**: `disposeSound()` cleans up the single stream's callbacks; `disposeAllSound()` cleans up all of them.
- **Cleanup on deinit**: `disposeNativeCallables()` (called during `deinit()` / hot-restart recovery) now also clears the callback map, preventing leaks across engine restarts.
- **Removed redundant dispose**: The old code called `_disposeBufferStreamCallbacks()` before inserting a new entry — this closed the previous callbacks before the map entry was replaced, which was unnecessary since map insertion handles replacement.

### Test

Added `BufferStreamCallbacks` on-device integration test (`example/tests/tests/`) that:
1. Creates a buffer stream with `onBuffering` callback
2. Plays before adding data to trigger the buffering → unbuffering transition
3. Verifies the callback fires
4. Disposes the stream (no crash)
5. Creates a second stream — verifies callbacks still work after prior disposal
6. Calls `disposeAllSources` — verifies bulk cleanup

**Note:** This PR should be merged after or together with #444 (hot-restart callback rebind), since `disposeNativeCallables` cleanup depends on the callback registration lifecycle introduced there.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/example/tests/tests/all_tests.dart
+++ b/example/tests/tests/all_tests.dart
@@ -3,6 +3,7 @@ import 'all_instances_finished.dart' as all_instances_finished;
 import 'async_multi_load.dart' as async_multi_load;
 import 'asynchronous_deinit.dart' as asynchronous_deinit;
 import 'auto_dispose.dart' as auto_dispose;
+import 'buffer_stream_callbacks.dart' as buffer_stream_callbacks;
 import 'buffer_stream_extended.dart' as buffer_stream_extended;
 import 'buffer_stream_small_mp3.dart' as buffer_stream_small_mp3;
 import 'compressor_filter.dart' as compressor_filter;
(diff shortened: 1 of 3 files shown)


alnitak:
Thanks you very much @skylartaylor, I was not aware of this problem!

Merging.

PR #466, opened 2026-05-17
feat(audio): expose master output capture

Slashpaf:
## Description

Adds an opt-in API for capturing the final SoLoud mixed output as interleaved float32 PCM.

The capture point is in the miniaudio backend immediately after `soloud->mix(...)`, before the buffer is submitted to the playback device. This gives applications access to the same final master mix that is being played by the engine.

The implementation does not touch existing code, only new implementation that does not affect flutter_soloud if not enabled. Web returns `notImplemented` (not using miniaudio).

Public Dart API added:
- `SoLoud.instance.setOutputCaptureEnabled(...)`
- `SoLoud.instance.readOutputCapture(maxFrames)`
- `SoLoud.instance.getOutputCaptureAvailableFrames()`
- `SoLoud.instance.getOutputCaptureDroppedFrames()`
- `SoLoud.instance.getOutputCaptureFormat()`

Use cases include local streaming, recording, remote monitoring, remote playback, integrations, and external visualization/analysis.

### DIFF
--- a/lib/src/bindings/bindings_player.dart
+++ b/lib/src/bindings/bindings_player.dart
@@ -985,6 +985,35 @@ abstract class FlutterSoLoud {
     bool average = false,
   });
 
+  // ///////////////////////////////////////
+  // output capture
+  // ///////////////////////////////////////
+
+  /// Enables or disables capture of the final mixed output.
+  ///
+  /// Captured samples are interleaved float32 PCM in the engine output format.
+  @mustBeOverridden
+  PlayerErrors setOutputCaptureEnabled(
+    bool enabled, {
+    int maxBufferedFrames = 44100,
+  });
+
+  /// Reads up to [maxFrames] captured frames.
+  ///
+  /// The returned list contains interleaved float32 samples. Its length is
+  /// `framesRead * channels`, where `channels` comes from
+  /// [getOutputCaptureFormat].
+  @mustBeOverridden
+  Float32List readOutputCapture(int maxFrames);
+
+  /// Returns captured frames currently available to read.
+  @mustBeOverridden
+  int getOutputCaptureAvailableFrames();
+
+  /// Returns the capture sample rate and channel count.
+  @mustBeOverridden
+  ({PlayerErrors error, int sampleRate, int channels}) getOutputCaptureFormat();
+
   /////////////////////////////////////////
   /// Mixing Bus
   /// https://solhsa.com/soloud/mixbus.html
--- a/lib/src/bindings/bindings_player_ffi.dart
+++ b/lib/src/bindings/bindings_player_ffi.dart
@@ -2388,6 +2388,113 @@ class FlutterSoLoudFfi extends FlutterSoLoud {
         )
       >();
 
+  // ///////////////////////////////////////
+  // output capture
+  // ///////////////////////////////////////
+
+  @override
+  PlayerErrors setOutputCaptureEnabled(
+    bool enabled, {
+    int maxBufferedFrames = 44100,
+  }) {
+    final error = _setOutputCaptureEnabled(enabled, maxBufferedFrames);
+    return PlayerErrors.values[error];
+  }
+
+  late final _setOutputCaptureEnabledPtr =
+      _lookup<
+        ffi.NativeFunction<ffi.UnsignedInt Function(ffi.Bool, ffi.UnsignedInt)>
+      >('setOutputCaptureEnabled');
+  late final _setOutputCaptureEnabled = _setOutputCaptureEnabledPtr
+      .asFunction<int Function(bool, int)>();
+
+  @override
+  Float32List readOutputCapture(int maxFrames) {
+    final format = getOutputCaptureFormat();
+    if (format.error != PlayerErrors.noError ||
+        format.channels <= 0 ||
+        maxFrames <= 0) {
+      return Float32List(0);
+    }
+
+    final out = calloc<ffi.Float>(maxFrames * format.channels);
+    final framesRead = calloc<ffi.UnsignedInt>();
+    try {
+      final error = _readOutputCapture(out, maxFrames, framesRead);
+      if (PlayerErrors.values[error] != PlayerErrors.noError ||
+          framesRead.value == 0) {
+        return Float32List(0);
+      }
+
+      final sampleCount = framesRead.value * format.channels;
+      return Float32List.fromList(out.asTypedList(sampleCount));
+    } finally {
+      calloc
+        ..free(out)
+        ..free(framesRead);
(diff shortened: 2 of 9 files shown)


Slashpaf:
@alnitak Just adding a concrete use case for this PR.

In Audio Forge I use this to stream the exact final SoLoud master mix to remote listeners. The app mixes music, ambience, effects, filters, volume changes, etc. locally, then reads the final interleaved float32 PCM output and sends it to a room server for redistribution.

This is different from capturing one source or using visualization data: the app gets the same final post-mix audio that is sent to the playback device(s).

Minimal usage:

```dart
final soloud = SoLoud.instance;

final format = soloud.getOutputCaptureFormat();

soloud.setOutputCaptureEnabled(true, maxBufferedFrames: 44100 * 2);

Timer.periodic(const Duration(milliseconds: 20), (_) {
  final available = soloud.getOutputCaptureAvailableFrames();
  if (available == 0) return;

  final samples = soloud.readOutputCapture(2048); // interleaved float32 PCM
  sendToServer(samples, format.sampleRate, format.channels);
});

// Later:
soloud.setOutputCaptureEnabled(false);

This enables use cases like remote playback, recording the full app mix, monitoring, broadcast rooms, and external audio analysis/visualization, while staying fully opt-in.

**Comment by alnitak** [comment]
Hi @Slashpaf, sorry for the long wait! I had to think about this because it is a tricky feature.

A while back, I already thought about implementing this feat, but there were some problems to take into account. I thought about a buffer too, but I'd preferred to have a stream for listening to the new audio data (instead of a `Timer.periodic` for checking).

This should be a nice feature to have, there is also an issue asking for this. Please let me think what we can do about this a little more! For example would be nice to have the audio data in other formats (instead of the default SoLoud interleaved floats32), or maybe already compressed audio with the OGG format.

**Comment by Slashpaf** [comment]
@alnitak no worries, thanks for the reply.

I agree that stream-based dart API would be nicer than the polls. This way users don't have to build polling loops.

I'll rework that.

For the first version I think keeping raw float32 PCM is best. It's the most flexible base format. Compressed output could be added later on top of this feature?

**Comment by alnitak** [comment]
> Compressed output could be added later on top of this feature?

Yes, I think that could be done later by setting some parameters when you enable capturing, and then filling the buffer with the format chosen.

Anyway, I don't want you to waste your time with something which maybe has a better solution. I think the tricky part is also to pass the data from C to Dart and do this in a way that doesn't interfere with the main UI isolate. A good way, I think, to read audio data from Flutter, is to use something like Filip wrote in his blog [here](https://filiph.net/text/making-my-1970s-renderer-multi-threaded.html) in the `FFI to the rescue` chapter. Doing that, I think it will also work for the web. For all this, I think the workflow of this PR should be reviewed using very few methods exposed to the user. Maybe just a `startCapture`, `stopCapture` and a stream listener to the isolate that works just to acquire new data (?).

Maybe this logic can also be used on the contrary: send audio data to the output mixer. I already saw a discussion for this, but there wasn't a (pro) solution due to the lack of Dart to share memory between isolates/threads (yet).

I also thought that the audio data acquired from Flutter could have something that identifies that chunk, like a sequence number, a timestamp, or a format.

Well, I need to draw some schematics and do some tests for this :).

**Comment by Slashpaf** [comment]
Okay, understood. I do require this feature so I went ahead and revised the implementation. The API is now much better with a simple outputCapture() stream method:

  ```dart
  final capture = SoLoud.instance.outputCapture();

  final subscription = capture.stream.listen((samples) {
    // interleaved float32 PCM
  });

  final sampleRate = capture.sampleRate;
  final channels = capture.channels;

  // Later:
  await subscription.cancel();
```

Much better to use now.

I also changed the native capture buffer so the audio callback path stays lock-free and does not block the audio thread.

The API should fit either implementation if you decide it should go another way.

Thanks again, I'll push the updated PR.

MaherSafadii:
what's the status on this PR? I just ran into an issue where audio played by the app doesn't get caught by screen recorders and it's a blocker for me.

alnitak:
Hey @MaherSafadii, I wish to add this PR or part of it. I still don't have much free time these days, but in July I think I will.

alnitak:
@Slashpaf, @MaherSafadii, I've implemented this feature in PR #495.

I'd really appreciate it if you could give it a try and share your feedback. Any comments or suggestions are very welcome and will help refine the implementation. Thank you!

Slashpaf:
@alnitak I just tested. It works, but I ran into an issue with the stream chunking for live audio.

It looks like `notificationThresholdBytes` means “notify when at least this much data is available”, but the emitted chunks are not guaranteed to be exactly that size. Is that intentional?

For recording or analysis this is probably fine, but for live network audio I need to turn the captured stream back into stable frame-sized packets before forwarding it. Without that packetization, the remote audio timing becomes unstable.

So the usage ends up looking like this:

```dart
final stream = SoLoud.instance.startMixerOutputStream(
  format: MixerOutputFormat.pcmS16le,
  sampleRate: 44100,
  channels: 2,
  bufferSizeBytes: maxBufferedFrames * 2 * 2,
  notificationThresholdBytes: chunkFrames * 2 * 2,
);

final pending = BytesBuilder(copy: false);

stream.listen((chunk) {
  pending.add(chunk);

  final bytes = pending.takeBytes();
  var offset = 0;
  final packetBytes = chunkFrames * 2 * 2;

  while (bytes.length - offset >= packetBytes) {
    final packet = Uint8List.sublistView(bytes, offset, offset + packetBytes);
    sendToRoomServer(packet);
    offset += packetBytes;
  }

  if (offset < bytes.length) {
    pending.add(Uint8List.sublistView(bytes, offset));
  }
});
```
With PR #466, where the app explicitly read fixed frame chunks, the usage was simpler:

```dart
final capture = SoLoud.instance.outputCapture(
  maxBufferedFrames: 44100 * 2,
  chunkFrames: 2048,
);

capture.stream.listen((samples) {
  final pcm16 = float32ToPcm16(samples);
  sendToRoomServer(pcm16);
});
```

Would it make sense for the API to optionally guarantee fixed-size chunks? For example:

```dart
SoLoud.instance.startMixerOutputStream(
  format: MixerOutputFormat.pcmS16le,
  sampleRate: 44100,
  channels: 2,
  chunkFrames: 2048,
);
```

where each stream event contains exactly chunkFrames * channels * bytesPerSample bytes, except maybe the final tail when capture stops.

alnitak:
Hey @Slashpaf, thanks for trying it! Could you please copy/paste your latest message to the PR, just to keep track of this for further readers and to interact directly within the PR? Thanks!

alnitak:
Closing in favor of #495

PR #479, opened 2026-06-02
[codex] Load Ogg Opus files through buffer stream fallback

Kunstderfug:
## What changed

This adds a native fallback in `Player::loadFile()` for Ogg Opus files. The normal SoLoud `Wav`/`WavStream` path is still tried first; if that fails and the file is detected as Ogg Opus, the file bytes are loaded into a preserved `BufferStream` using the existing auto decoder.

## Why

`loadFile()` already works for Ogg Vorbis and FLAC through SoLoud, but Ogg Opus files can fail through the `WavStream` path even though the package has an Opus decoder in the buffer-stream pipeline. This lets callers use `loadFile()` directly for Opus files instead of pre-decoding them to temporary WAV files.

## Impact

Ogg Opus files can now be loaded, played, and seeked through the native player path. Other formats continue to use the existing `Wav`/`WavStream` behavior.

## Validation

- `flutter analyze`
- `flutter test`
- `flutter build macos --debug` in `example/`
- Direct native smoke against a real `.opus` file: `loadFile` returned `noError`, paused `play` returned `noError`, `seek(12.5)` returned `noError`, and `getPosition()` reported `12.500`.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -27,7 +27,7 @@
 #endif
 
 namespace {
-constexpr unsigned int kOggOpusBufferStreamMaxBytes = 512u * 1024u * 1024u;
+constexpr unsigned int kOggXiphBufferStreamMaxBytes = 512u * 1024u * 1024u;
 
 bool readFileBytes(const std::string &filePath,
                    std::vector<unsigned char> &bytes)
@@ -48,7 +48,7 @@ bool readFileBytes(const std::string &filePath,
     return file.gcount() == fileSize;
 }
 
-bool isOggOpusBytes(const std::vector<unsigned char> &bytes)
+bool isOggXiphBytes(const std::vector<unsigned char> &bytes)
 {
     if (bytes.size() < 35 || std::memcmp(bytes.data(), "OggS", 4) != 0) {
         return false;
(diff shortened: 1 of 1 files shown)


alnitak:
Hi @Kunstderfug, sorry for the wait.

Since the Ogg container also supports the FLAC codec, I have added a check for that also.

To recap: 
1. the Vorbis codec is managed by `stb_vorbis.c`
2. Flac and Opus are managed by Buffer stream with this PR

By using `loadfile()` with 2, it makes impractical the use of its `mode: LoadMode.disk` parameter since the Buffer stream does decode the whole audio into memory.

I will merge this, but it would be nice in the future to get rid of `stb_vorbis.c` and use the Xiph libraries for the Ogg format directly in [soloud_wav.cpp]( https://github.com/alnitak/flutter_soloud/blob/main/src/soloud/src/audiosource/wav/soloud_wav.cpp) and in [soloud_wavstream.cpp](https://github.com/alnitak/flutter_soloud/blob/main/src/soloud/src/audiosource/wav/soloud_wavstream.cpp).

Thanks again for your valuable support!!

alnitak:
Hey @Kunstderfug, in the `stbvorbisRid` I am using Xiph libraries to get rid of `stb_vorbis.c`. This lets you open also OGGs files with Opus and FLAC formats using the standard `load*` methods without using the `setBufferStream` fallback. And you can also use the `mode: LoadMode.disk`. 
Also, there was a bug in `setBufferStream` which couldn't load OGG/FLAC files. This is fixed and it could also read plain FLAC files.

If you wish to try, please change your dep in `pubspec.yaml` like this:

```
dependencies:
  flutter_soloud:
    git:
      url: https://github.com/alnitak/flutter_soloud.git
      ref: stbvorbisRid
```

Thanks.

PR #488, opened 2026-06-21
Fix web: don't spawn the deferred-pause std::thread on web (Aborted() in initEngine)

felixmin:
## Description

On web, `SoLoud.init()` aborts with `RuntimeError: Aborted()` inside `initEngine` as soon as the engine initializes — a regression in 4.0.10.

4.0.10 (#486) starts a background pause-scheduler thread in `Player::init()` (`startPauseEngineScheduler()` → `std::thread`). The web build is single-threaded (`web/compile_wasm.sh` passes no `-pthread`), so constructing `std::thread` calls `pthread_create` with no thread runtime and the module aborts on the first `init()`. 4.0.9 had no such thread.

This guards the thread behind `__EMSCRIPTEN__`:

- `startPauseEngineScheduler()` / `stopPauseEngineScheduler()` are no-ops on web (no `std::thread` is constructed).
- `pauseEngine()` falls back to the pre-4.0.10 immediate pause on web — the browser's AudioContext doesn't have the OS audio-session settling issue that motivates the delay on native.

Native (iOS/Android/desktop) keeps the deferred, coalesced pause unchanged.

The web artifacts (`web/libflutter_soloud_plugin.{js,wasm}`) are regenerated via `web/compile_wasm.sh` (Emscripten 6.0.0, with Opus/Ogg) and committed.

**Verified** in Chrome with the rebuilt artifacts by calling `initEngine`:

- published 4.0.10 → `Aborted(). Build with -sASSERTIONS for more info.` (the crash)
- this branch → `initEngine` returns `0`, `isInited()` returns `1` — no abort

> Note: the wasm here was built with emsdk 6.0.0. If you release with a different Emscripten version, regenerating with yours will keep the committed binary aligned with your toolchain — happy to update if you prefer.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felixmin:
### Runtime verification (real browser)

I rebuilt the web wasm from this branch with Emscripten 6.0.0 and ran `initEngine` directly in Chrome, comparing it against the published 4.0.10 artifacts. Minimal harness: load the module factory, then `Module.ccall('initEngine', 'number', ['number','number','number','number'], [-1, 44100, 2048, 2])`.

| Build | Result |
|---|---|
| **4.0.10 (published `web/libflutter_soloud_plugin.*`)** | `Aborted(). Build with -sASSERTIONS for more info.` — i.e. the reported crash, reproduced |
| **This branch (rebuilt wasm)** | `initEngine` returns `0` (noError), `isInited()` returns `1` — no abort |

So the `__EMSCRIPTEN__` guard resolves the web init crash. (The patched test build was compiled without the Xiph/opus libs for speed, but that's irrelevant to this code path — the abort is solely the `std::thread` constructed in `Player::init()`, which the guard removes on web.)

Still source-only here — the committed `web/libflutter_soloud_plugin.{js,wasm}` should be regenerated with your pinned Emscripten toolchain so the artifact diff matches your setup. Happy to push a rebuild if you share the emsdk version you release with.

alnitak:
Thank you so much! LGTM.

> Happy to push a rebuild if you share the emsdk version you release with.

Thanks, I will rebuild wasm before publishing on pub.dev.

PR #492, opened 2026-06-27
add `lowLatency` init option to allow recordable Android output

MjnMixael:
## Description

`init()` always opened the playback device with miniaudio's default low-latency performance profile. On Android that maps to AAUDIO_PERFORMANCE_MODE_LOW_LATENCY and takes AAudio's MMAP/exclusive path, which has two surprising consequences:

  1. It bypasses the AudioFlinger mixer that system screen recorders tap, so app audio is silent in screen recordings.
  2. Its tiny, tight-deadline callbacks starve CPU-heavy filters (such as the FFT pitch shift), producing audible dropouts.

This adds an opt-in `bool lowLatency = true` parameter to `init()`. The default preserves the existing behavior on every platform, so this is hopefully non-breaking. When set to `false`, the device is opened with the conservative (legacy mixer) profile; on Android the stream is also tagged MEDIA/MUSIC with capture explicitly allowed. The result is output that screen recorders can capture and that gives heavy DSP enough headroom to run cleanly, at the cost of higher output latency.

The flag is threaded init() -> initEngine() (FFI) -> Player::init() -> SoLoud::miniaudio_setLowLatency() and consumed in miniaudio_init(), avoiding any change to the vendored SoLoud core's init() signature. The Web backend accepts and ignores the flag.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
+++ b/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
@@ -45,6 +45,25 @@ namespace SoLoud
     {
         gMiniaudioLowLatency = aLowLatency;
     }
+
+    // Android (AAudio) stream attributes applied when low-latency is disabled.
+    // Default to media/music (sensible for a media app, and capturable). When
+    // [aManaged] is false the app wants to own AudioAttributes externally (e.g.
+    // via the audio_session plugin), so we leave usage/contentType unset
+    // (`_default`) and let AAudio pick its defaults. Stored in globals so the
+    // SAME choice is re-applied on device changes (see both call sites) rather
+    // than reverting. Defined outside the WITH_MINIAUDIO guard so the setter
+    // symbol always exists for the C bindings to call.
+    static ma_aaudio_usage gMiniaudioAaudioUsage = ma_aaudio_usage_media;
+    static ma_aaudio_content_type gMiniaudioAaudioContentType = ma_aaudio_content_type_music;
+
+    void miniaudio_setAndroidAudioAttributes(bool aManaged)
+    {
+        gMiniaudioAaudioUsage =
+            aManaged ? ma_aaudio_usage_media : ma_aaudio_usage_default;
+        gMiniaudioAaudioContentType =
+            aManaged ? ma_aaudio_content_type_music : ma_aaudio_content_type_default;
+    }
 }
 
 #if !defined(WITH_MINIAUDIO)
@@ -404,12 +423,17 @@ namespace SoLoud
         
 #elif defined(__ANDROID__)
         // When low-latency is disabled the device runs on the legacy mixer path
-        // (set above). Also tag the AAudio stream as media/music and explicitly
-        // allow capture so system screen recorders pick up the audio.
+        // (set above). Tag the AAudio stream with the configured usage/contentType
+        // (media/music by default; left unset when the app opts to manage
+        // AudioAttributes externally via audio_session — see
+        // miniaudio_setAndroidAudioAttributes) and explicitly allow capture so
+        // system screen recorders pick up the audio. miniaudio skips the setter
+        // for `_default`, so opting out truly leaves the attributes untouched.
+        // The same globals are re-applied on device changes (changeDevice_impl).
         if (!gMiniaudioLowLatency)
         {
-            deviceConfig.aaudio.usage                = ma_aaudio_usage_media;
-            deviceConfig.aaudio.contentType          = ma_aaudio_content_type_music;
+            deviceConfig.aaudio.usage                = gMiniaudioAaudioUsage;
+            deviceConfig.aaudio.contentType          = gMiniaudioAaudioContentType;
             deviceConfig.aaudio.allowedCapturePolicy = ma_aaudio_allow_capture_by_all;
         }
 
@@ -560,8 +584,14 @@ namespace SoLoud
 #if defined(__ANDROID__)
         if (!gMiniaudioLowLatency)
         {
-            deviceConfig.aaudio.usage                = ma_aaudio_usage_media;
-            deviceConfig.aaudio.contentType          = ma_aaudio_content_type_music;
+            // Re-apply the SAME attributes chosen at init so a device change
+            // doesn't silently revert them. If the app opted out
+            // (miniaudio_setAndroidAudioAttributes(false)), these are `_default`
+            // and miniaudio leaves them unset — so an externally-managed
+            // configuration (e.g. via audio_session) is preserved across device
+            // changes rather than being forced back to media/music.
+            deviceConfig.aaudio.usage                = gMiniaudioAaudioUsage;
+            deviceConfig.aaudio.contentType          = gMiniaudioAaudioContentType;
             deviceConfig.aaudio.allowedCapturePolicy = ma_aaudio_allow_capture_by_all;
         }
 #endif
--- a/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
+++ b/src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
@@ -30,6 +30,23 @@ distribution.
 #include <unistd.h>
 #endif
 
+namespace SoLoud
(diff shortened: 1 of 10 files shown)


src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
```diff
@@ -379,6 +403,16 @@ namespace SoLoud
         gDeviceStartDeferred = false;
         
 #elif defined(__ANDROID__)
+        // When low-latency is disabled the device runs on the legacy mixer path
+        // (set above). Also tag the AAudio stream as media/music and explicitly
+        // allow capture so system screen recorders pick up the audio.
+        if (!gMiniaudioLowLatency)
+        {
+            deviceConfig.aaudio.usage                = ma_aaudio_usage_media;
+            deviceConfig.aaudio.contentType          = ma_aaudio_content_type_music;
```
alnitak: I'm thinking of users who use the [`audio_session`](https://pub.dev/packages/audio_session) plugin with which it is also possible to set these 2 values. The author says in the [doc](https://pub.dev/packages/audio_session#configuring-the-audio-session):

> t is recommended that you apply your own preferred configuration using audio_session after all other audio plugins have loaded.

but he says also:

> You may consider asking the developer of each audio plugin you use to provide an option to not overwrite these global settings and allow them be managed externally.

If setting these parameters is mandatory, we can then write a warning in the CHANGELOG.md or maybe in the doc of the `init` method. If they are not mandatory, but suggested, we can write this only in the doc of the `init` method.

What do you think?

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
```diff
@@ -517,6 +551,21 @@ namespace SoLoud
...
+        // otherwise switching the output device would silently revert to the
+        // default low-latency/MMAP path (see gMiniaudioLowLatency).
+        deviceConfig.performanceProfile = gMiniaudioLowLatency
+            ? ma_performance_profile_low_latency
+            : ma_performance_profile_conservative;
+#if defined(__ANDROID__)
+        if (!gMiniaudioLowLatency)
+        {
+            deviceConfig.aaudio.usage                = ma_aaudio_usage_media;
+            deviceConfig.aaudio.contentType          = ma_aaudio_content_type_music;
```
alnitak: Same as above

alnitak:
Hi @MjnMixael, thanks a lot for your PR! Much appreciated.

I have just one question in the review.

MjnMixael:
> Hi @MjnMixael, thanks a lot for your PR! Much appreciated.
> 
> I have just one question in the review.

Happy to answer any questions. Just let me know!

alnitak:
I have written the question in the review [here](https://github.com/alnitak/flutter_soloud/pull/492/changes/BASE..7ac3f8b010bfaaaeba5e64e69d21b009a79757db#r3497371747) :)

It's just a nit

MjnMixael:
> I have written the question in the review [here](https://github.com/alnitak/flutter_soloud/pull/492/changes/BASE..7ac3f8b010bfaaaeba5e64e69d21b009a79757db#r3497371747) :)
> 
> It's just a nit

It may be me but that link just takes me to the files and the PR doesn't show any reviews yet. Any chance you accidentally forgot to save and post the review?

alnitak:
yes sorry! Forgot to save!

alnitak (commented):
here

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
MjnMixael: Good question. These aren't mandatory. Playback works without them; they only hint Android's routing/ducking/volume behavior, so media/music are just sensible defaults. I'll document this in the init method docs. I can also expose them (defaulting to media/music, with a _default value to leave them untouched) so anyone managing attributes via audio_session can opt out. I'm happy to add that here or as a follow-up, whichever you prefer.

alnitak: IMHO, defaulting them and writing something in the init doc could be the right solution. Something like that, using the `lowLatency` parameter set to false, is preferred to set those 2 conf parameters to `media` and `music` by using the `audio_session` plugin. But it's up to you because I am not so prepared on this kind of stuff!

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
alnitak: Re-thinking what we wrote above, the code here is called in some situations during the app run. So if the user has chosen some other configs, by setting these 2 to a default value, will overwrite what the user chose with `audio_session`. Maybe these 2 could be removed?

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
MjnMixael: Done.I exposed it rather than leaving it implicit. init() now takes an androidAudioAttributes parameter (AndroidAudioAttributes enum):

mediaMusic (default) → tags the stream usage = media / contentType = music, same as today. Sensible for a media app and capturable by screen recorders.
unmanaged → leaves usage/contentType unset so a caller managing AudioAttributes via audio_session (or otherwise) isn't overridden; the capture policy is still applied.
I documented it on init(), including that these are the stream's attributes (audio_session controls the focus request. For correct ducking they should match) and that it's a no-op with low-latency enabled / off-Android / on web. Since it's opt-in via the param with unchanged defaults, I don't think a CHANGELOG warning is needed, just the init doc.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
MjnMixael: Good catch that this runs during the app's lifetime. I went a slightly different way than removing them, because dropping them here would make a device change revert the stream to AAudio's defaults and discard whatever the caller chose at init. So both init and this path now apply the same stored choice (globals set by miniaudio_setAndroidAudioAttributes):

Default (mediaMusic) → re-applies media/music on device change, matching init.
Opt-out (unmanaged) → the globals are ma_aaudio_usage_default / ma_aaudio_content_type_default, and miniaudio skips the setter for _default, so the attributes are genuinely left untouched across device changes too, nothing is overwritten.
One nuance worth noting: audio_session configures the app's focus request AudioAttributes at the AudioManager level... it doesn't set the per-AAudio-stream usage/contentType that miniaudio creates. So leaving them unset here doesn't make the stream inherit audio_session's values; it falls back to AAudio's defaults. The new param is what lets a caller align the stream's attributes with their focus request. This keeps the choice (set or opt-out) consistent whether or not the output device changes.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
alnitak: Good! Thank you very much.

alnitak:
Thank you very much! I was not aware of the low latency parameter.

I will merge and I'll update on pub.dev soon.

src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp
alnitak: Ah ok. Good choice.

Also, this PR let me think that, in the near future, I will add some more configurations also for the other platforms ;)

MjnMixael: Awesome! Thanks for the merge and thanks for your work on this plugin. It's very useful!

alnitak:
Hey @MjnMixael, just to let you know. I had to rename the `enum AndroidAudioAttributes` to `AndroidAAudioAttributes` (doubled `A`) because it was conflicting with the `AndroidAudioAttributes` enum of `audio_session` plugin. Same for all the other functions and parameters like `setAndroidAAudioAttributes`.

Thanks again.

PR #495, opened 2026-07-03
capture the master mixer output as a `Stream<Uint8List>`

### LINKED ISSUE #437: feat: Ability to listen to engine output
My app is a soundscape mixer that I'd like to be able to export audio like DAWs do. On iOS there is no way to record the internal audio in an app, so currently if I want users to be able to export/record the app's sound output to a single audio file the only options are to:
- tell them to screen record and extract the audio (which isn't lossless), or
- separately render the audio with ffmpeg (which would likely be inconsistent when using filters, oscillations, etc)

If flutter_soloud exposed a way to listen to the engine's PCM data that would solve this problem (unless there's another solution or way to do it that I don't know about). I'm not sure how easy this would be, maybe it is out of the scope of the package.

alnitak:
## Description

- added `autoDispose` parameter to `setBufferStream` to automatically dispose the sound when it is finished. This eliminates the need to call disposeSource manually.
- **added mixer output capture**: capture the master mixer output as a `Stream<Uint8List>` in PCM (F32LE, S8, S16LE, S32LE) or compressed formats (Opus, Vorbis, FLAC, WAV). See `SoLoud.startMixerOutputStream` / `stopMixerOutputStream` / `isMixerOutputStreamRunning`.
- **added `mixer_capture` example** (`example/lib/mixer_capture/mixer_capture.dart`) that shows how to capture the master mix and save it to a file.
- **web note**: the web build requires `--wasm` (or any server that sends `Cross-Origin-Opener-Policy: same-origin` and `Cross-Origin-Embedder-Policy: require-corp` headers) because the WASM module uses `SharedArrayBuffer` for the audio thread. Running with `flutter run -d chrome` without `--wasm` is not supported by the default dev server.

New doc available [here](https://docs.page/alnitak/flutter_soloud_docs~dev/advanced/mixer_output_capture).

closes #437 issue
closes #466 PR

## Type of Change


Slashpaf:
@alnitak I just tested. It works, but I ran into an issue with the stream chunking for live audio.

It looks like `notificationThresholdBytes` means “notify when at least this much data is available”, but the emitted chunks are not guaranteed to be exactly that size. Is that intentional?

For recording or analysis this is probably fine, but for live network audio I need to turn the captured stream back into stable frame-sized packets before forwarding it. Without that packetization, the remote audio timing becomes unstable.

So the usage ends up looking like this:

```dart
final stream = SoLoud.instance.startMixerOutputStream(
  format: MixerOutputFormat.pcmS16le,
  sampleRate: 44100,
  channels: 2,
  bufferSizeBytes: maxBufferedFrames * 2 * 2,
  notificationThresholdBytes: chunkFrames * 2 * 2,
);

final pending = BytesBuilder(copy: false);

stream.listen((chunk) {
  pending.add(chunk);

  final bytes = pending.takeBytes();
  var offset = 0;
  final packetBytes = chunkFrames * 2 * 2;

  while (bytes.length - offset >= packetBytes) {
    final packet = Uint8List.sublistView(bytes, offset, offset + packetBytes);
    sendToRoomServer(packet);
    offset += packetBytes;
  }

  if (offset < bytes.length) {
    pending.add(Uint8List.sublistView(bytes, offset));
  }
});
```
With PR #466, where the app explicitly read fixed frame chunks, the usage was simpler:

```dart
final capture = SoLoud.instance.outputCapture(
  maxBufferedFrames: 44100 * 2,
  chunkFrames: 2048,
);

capture.stream.listen((samples) {
  final pcm16 = float32ToPcm16(samples);
  sendToRoomServer(pcm16);
});
```

Would it make sense for the API to optionally guarantee fixed-size chunks? For example:

```dart
SoLoud.instance.startMixerOutputStream(
  format: MixerOutputFormat.pcmS16le,
  sampleRate: 44100,
  channels: 2,
  chunkFrames: 2048,
);
```

where each stream event contains exactly chunkFrames * channels * bytesPerSample bytes, except maybe the final tail when capture stops.

alnitak:
@Slashpaf thanks for the quick feedback!

You’re right: `notificationThresholdBytes` is a minimum threshold, not a fixed chunk size. The native side emits a contiguous slice from the circular buffer whenever at least that many bytes are available, so the actual chunk size can vary depending on how the audio callback and notification thread line up.

That behavior is intentional because it keeps the audio-thread path simple and lock-free, but I completely understand that for live network audio, fixed-size frames are much easier to consume.

I think the cleanest way to support both is to add an optional `chunkFrames` parameter to `startMixerOutputStream`. When set, the native side would buffer internally and only emit once it has exactly `chunkFrames * channels * bytesPerSample` bytes, dropping/padding the tail only on stop. When not set, it keeps the current threshold-based behavior for recording/analysis use cases. But this new `chunkFrames` parameter would only apply to PCM formats. So, maybe better call it `chunkPCMFrames` :)

If that API works for your network audio case, I think it can be done. WDYT?

Slashpaf:
Sounds good!

alnitak:
Hey @Slashpaf. I left the`notificationThresholdBytes` parameter and added `chunkPCMFrames`. Here is what the doc says:

| | |
|-----------|-------------|
| `notificationThresholdBytes` | Number of bytes that must be available before a chunk is emitted. Used for compressed formats and for PCM when `chunkPCMFrames` is `-1`. Smaller values give lower latency but more callbacks. |
| `chunkPCMFrames` | Fixed number of PCM frames per emitted chunk. PCM only; must be at least `2048` and a multiple of `channels * bytesPerSample`. `-1` disables fixed-size chunking. |

Slashpaf:
@alnitak Thank you, `chunkPCMFrames` sounds exactly like what I need.

Small thing: I think the validation/doc mixes frames and bytes. Since `chunkPCMFrames` is a frame count, it should not need to be a multiple of `channels * bytesPerSample`.

The emitted byte size would simply be:

```dart
chunkPCMFrames * channels * bytesPerSample
```

So maybe the constraint should just be:

> PCM only; must be -1 or at least 2048.

---

I did some tests and it seems that it still has issues for live audio.

The chunk size is now correct. With:

```dart
chunkPCMFrames: 2048,
format: MixerOutputFormat.pcmS16le,
channels: 2,
```

I consistently receive 8192 byte chunks, which is correct:

2048 * 2 * 2

But I still get capture gaps of around 1 second during normal app activity, especially when loading/creating audio players. I added temporary diagnostics locally and saw this pattern:

[SOLOUD_CAPTURE] native fixed chunk stalled: stalledFor=250ms available=49152 chunkBytes=8192
[SOLOUD_CAPTURE] native fixed chunk stalled: stalledFor=500ms available=90112 chunkBytes=8192
[SOLOUD_CAPTURE] native fixed chunk stalled: stalledFor=751ms available=139264 chunkBytes=8192
[SOLOUD_CAPTURE] Dart callback delayed: gap=1007ms length=8192 availableBefore=172032 availableAfter=163840
[MasterCapture] capture anomaly: capture gap 1008ms, chunkBytes=8192

So native has audio available, but capture delivery appears to stall while waiting for Dart to process/advance the previous chunk. That makes the live stream unstable even though the fixed chunk size itself is correct.

Maybe fixed PCM mode needs to avoid a single in-flight reusable chunk buffer? For live audio, native probably needs to keep queuing/capturing fixed chunks even if Dart is briefly busy, and let Dart drain them afterward.

So I think chunkPCMFrames is the right API direction, but the delivery path still seems coupled to Dart callback responsiveness.

Slashpaf:
One more thing I forgot to mention: before testing the callback-delay issue above, I also hit a separate fixed-chunk notification issue.

In fixed PCM mode, `m_notified` was only reset when available data dropped below the threshold/chunk size. During continuous playback, available data often never drops below one chunk, so native would emit one `8192` byte chunk and then wait until the ring buffer state changed much later.

That produced very regular ~2 second capture gaps for me, matching my `maxBufferedFrames = 44100 * 2` buffer duration.

Locally I changed `advanceReadPosition()` from resetting `m_notified` only when `available < threshold` to resetting it whenever `toAdvance > 0`:

```cpp
if (toAdvance > 0) {
    m_notified.store(false, std::memory_order_release);
}
```

That fixed the regular ~2s gaps, but the remaining ~1s gaps still happen when Dart is delayed, as described above.

alnitak:
> it should not need to be a multiple of `channels * bytesPerSample`.

Oh yes, thanks to finding that!

Anyway, I think that dealing with the single threading of Dart, is the main problem of this kind of problem, especially when doing audio. The audio chunks are delivered synchronously to Dart from native, and if Dart is doing something else, you could notice gaps. 

You could try to put the start, stop capturing and the listener into an Isolate (I think it can be done without initializing SoLoud because the audio engine is in another thread), and I am sure you will get a better result.

You are right about checking `if (toAdvance > 0)`. I'll fix it.

If you have a minimal test code to share, it will be great!

Thanks for testing!

Slashpaf:
@alnitak Yes, I understand the Dart isolate concern, and an isolate may help.

But the reason I’m concerned is that my previous PR worked reliably in the same app, during the same loading/player creation work. The difference is that it was pull-based: native kept filling a ring buffer, and Dart could read accumulated frames after a short delay.

With the current push callback path, native seems to have one in-flight chunk and waits for Dart to copy/advance it. When Dart is busy, capture delivery stalls even though native has audio available.

So I think an isolate can be a workaround, but the stream implementation probably still needs native-side buffering/queueing so capture is not coupled to Dart callback responsiveness.

I’ll test with an isolate though, and try to share a minimal repro.

Slashpaf:
@alnitak I tested the isolate idea, but it does not work with the current API.

Minimal test:

```dart
void captureIsolateMain(SendPort replyPort) {
  try {
    SoLoud.instance.startMixerOutputStream(
      format: MixerOutputFormat.pcmS16le,
      sampleRate: 44100,
      channels: 2,
      chunkPCMFrames: 2048,
    ).listen((chunk) {
      replyPort.send(chunk.length);
    });
  } catch (error) {
    replyPort.send(error.toString());
  }
}
```

Result:

`SoLoudNotInitializedException: SoLoud has not been initialized yet.`

So the listener cannot simply be moved to another isolate, because that isolate has its own Dart SoLoud.instance state. Am I missing something?

For comparison, my previous implementation worked reliably because native kept writing to a ring buffer and Dart only pulled fixed-size chunks from it:

```dart
final capture = SoLoud.instance.outputCapture();

capture.stream.listen((samples) {
  sendToRoomServer(samples);
});
```

That was much simpler to use and more robust in my tests, because capture did not depend on Dart immediately handling each native callback.

alnitak:
> So the listener cannot simply be moved to another isolate, because that isolate has its own Dart SoLoud.instance state. Am I missing something?

oh yes, sorry! For a few versions, I've added a check for the temporary cache initialization which causes the problem. But probably, re-initializing soloud doesn't hurt (or maybe remove [this](https://github.com/alnitak/flutter_soloud/blob/a99f90fc3364baa70d19d2ff3556cc4f65a76600/lib/src/soloud.dart#L284) last check).

To recap:
- we know that this could work (as your PR)
- a buffer queue of chunks, in native, could solve

If you can provide a simple example that prints out a log similar to the one you posted before:
> [SOLOUD_CAPTURE] native fixed chunk stalled: stalledFor=751ms available=139264 chunkBytes=8192
> [SOLOUD_CAPTURE] Dart callback delayed: gap=1007ms length=8192 availableBefore=172032 availableAfter=163840
> [MasterCapture] capture anomaly: capture gap 1008ms, chunkBytes=8192

alnitak:
I have added an [example](https://github.com/alnitak/flutter_soloud/blob/mixerOutput/example/lib/mixer_capture/isolate_capture_test.dart) to run the capture from another isolate other than the main.
The trick is this:
```dart
    final controller = SoLoudController();
    final bindings = controller.soLoudFFI;

    await bindings.setDartEventCallbacks();
```

then you can use the `bindings` as you use `SoLoud.instance`.

While the capture isolate is running, you can test to block the main UI and you will see that the stream is continuing to receive chunks.

Also, now if you run the mixing capture from the main isolate, chunks are still queued for "better time" to be delivered. 

Hope it works for you also!

Slashpaf:
I see your example works by using the internal FFI bindings directly.

That makes sense for testing, but I think it means isolate capture is not really exposed through the public SoLoud.instance API yet.

One thing I am still unsure about in the new native queue change: in fixed chunk mode the native side now advances the read position before Dart receives/copies the chunk. Is the memory passed to Dart guaranteed to stay valid until Dart copies it?

My concern is that if the callback receives a pointer to a reusable native chunk buffer, and Dart is delayed, native may overwrite that buffer before Dart actually copies it.

That would explain the kind of audio corruption/dropouts I was hearing before.

I’ll test the latest version again, but I think the robust version probably needs each emitted chunk to own stable memory until Dart has copied it, or to queue copied chunks natively.

alnitak:
@Slashpaf thanks again for the deep testing.

> That makes sense for testing, but I think it means isolate capture is not really exposed through the public SoLoud.instance API yet.

you're right that it uses the internal SoLoudController / FFI bindings directly. Moving the listener to a public-API isolate is something we can expose later, but the engine itself is a single C++ singleton, so the current way to run capture off the main thread is through that low-level controller. The example is mainly there to prove the concept.

> My concern is that if the callback receives a pointer to a reusable native chunk buffer, and Dart is delayed, native may overwrite that buffer before Dart actually copies it.

the callback is synchronous. The native notification thread does not reuse `m_chunkBuffer` until the Dart callback returns. Inside the Dart callback we immediately copy the bytes into a `Uint8List` before adding them to the stream: [native](https://github.com/alnitak/flutter_soloud/blob/2f688b3337ce5de7552559f0c38eb83c53d221da/src/mixeroutput/mixer_output.cpp#L337-L349) -> [Flutter](https://github.com/alnitak/flutter_soloud/blob/2f688b3337ce5de7552559f0c38eb83c53d221da/lib/src/bindings/bindings_player_ffi.dart#L206-L207).
The public StreamController is async only to schedule listener delivery; the copy itself happens while the native callback is still running, so the native buffer is not in danger of being overwritten.

Please try the latest commit and let me know if you still hear something bad. If it does, I could add a per-chunk queue in native so Dart can fall behind without stalling the capture thread, but I think that this change will be harder to do.

Slashpaf:
@alnitak I tested more. The big remaining issue was not capture once it ran in an isolate, but that my app was still forwarding chunks back to the main isolate before sending them over WebSocket.

When I moved the WebSocket connection and binary audio send into the same capture isolate, the drops disappeared. So I think all is good.

This works with the internal `SoLoudController` approach, but yes, it would be much nicer if the plugin exposed a public supported way to set up mixer output capture from a worker isolate :).

alnitak:
@Slashpaf that's great!

I'll merge this to the `dev` branch for now.

I will do another PR, and cc you if you wish, to write a new class that manages the `SoLoudController` and exposes methods that can be run out of the main isolate, like the `startMixerOutputStream` (still figuring out if this could be a good way of doing this).

Thanks again for your time spent helping me :)

PR #499, opened 2026-07-13
feat: add native loop end points

### LINKED ISSUE #498: feat: add native loop end points
**Description**

Add a native end point for loop regions. flutter_soloud already exposes a loop start through `loopingStartAt` and `getLoopPoint` / `setLoopPoint`, but looping always wraps at the natural end of the source. This makes sample-accurate regions inside a longer source impossible.

The end point needs to be enforced by the native mixer while it fills source buffers. Dart-side timers or post-buffer checks can overshoot the requested boundary.

**Requirements**

- [ ] Add an optional `loopingEndAt` parameter alongside `loopingStartAt` on the regular, 3D, source, and bus playback APIs.
- [ ] Add live-handle `getLoopEndPoint` and `setLoopEndPoint` APIs; `null` clears the end point and restores natural-EOF looping.
- [ ] Treat a region as half-open: `[loop start, loop end)`.
- [ ] Reject negative start values and end values that are not later than the start.
- [ ] Enforce the boundary in native source-frame reads without overshooting a mixer block, including the inaudible-tick path.
- [ ] If an end point is beyond the source duration, keep the existing natural-EOF behavior.
- [ ] Preserve current first-pass behavior for each playback API and require a source that can seek back to the loop start.
- [ ] Cover short regions, exact buffer boundaries, natural EOF fallback, clearing the end point, and invalid ranges with deterministic tests.
- [ ] Keep FFI and web/WASM APIs in parity and document the new contract.

**Additional Context**

The missing end point was discussed while loop-start support was introduced in #43, and the underlying SoLoud limitation was also called out in jarikomppa/soloud#15. The implementation should live at the generic mixer seam so it works consistently across supported source types rather than only for one decoder.

Kunstderfug:
Closes #498

## Summary

- adds native per-voice half-open [start, end) loop regions for audible and inaudible voices
- exposes loopingEndAt plus live getLoopEndPoint and setLoopEndPoint APIs through Dart, native FFI, the SoLoud C API, and web bindings
- preserves the existing play and play3d ABI with additive loop-point entry points
- regenerates the web artifacts with the repository-compatible Emscripten 4.0.22 toolchain
- adds deterministic boundary, EOF, precision, seek, no-progress, and live-refill regression coverage

## Live update behavior

Getters reflect requested loop bounds immediately. Playback applies a live start/end change at the next 512-frame source refill, so already-decoded frames plus backend latency may still use the prior region. Bounds passed to play or play3d are installed before the first decoded sample.

## Verification

- native loop-region suite: 1,643 assertions passed
- existing limiter suite: 18 assertions passed
- Flutter tests passed
- native CMake build passed
- C API syntax check passed
- Apple unity translation unit check passed
- macOS example debug build passed
- WebAssembly example build passed
- package publish dry run passed with 0 warnings
- flutter analyze reports only 20 pre-existing experimental mixer API warnings


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - **web note**: the web build requires `--wasm` (or any server that sends `Cross-Origin-Opener-Policy: same-origin` and `Cross-Origin-Embedder-Policy: require-corp` headers) because the WASM module uses `SharedArrayBuffer` for the audio thread. Running with `flutter run -d chrome` without `--wasm` is not supported by the default dev server.
 - **added `SoLoudIsolate`**, an isolate-safe singleton for running mixer output capture (and other safe operations, for now only `readSamplesFrom*`) from a non-main isolate without touching the main isolate's loader, filters, or event callbacks.
 - added `example/lib/mixer_capture/isolate_capture_test.dart` to demostrate mixer output capture from a separate isolate.
+- added native loop end points through `loopingEndAt` and the live `getLoopEndPoint` / `setLoopEndPoint` APIs, allowing half-open `[start, end)` loop regions.
```
alnitak:
```suggestion
- added native loop end points through `loopingEndAt` and the live `getLoopEndPoint` / `setLoopEndPoint` APIs, allowing half-open `[start, end)` loop regions #499. Thanks to @Kunstderfug 
```

src/soloud/src/core/soloud_core_looping.cpp
```diff
@@ -0,0 +1,111 @@
...
+   in a product, an acknowledgment in the product documentation would be
+   appreciated but is not required.
+
+   2. Altered source versions must be plainly marked as such, and must not be
+   misrepresented as being the original software.
+
+   3. This notice may not be removed or altered from any source
+   distribution.
+*/
+
```
alnitak, line 23: I would remove this, or maybe change
```
SoLoud audio engine
Copyright (c) 2013-2020 Jari Komppa
```
to
```
Copyright (c) 2023-2026 the flutter_soloud authors
```

alnitak (changes requested):
Thanks a lot @Kunstderfug!

This is a great feature that could lead to support [`SoundFonts`](https://en.wikipedia.org/wiki/SoundFont)!

LGTM, just 2 nit.

Issue #481, opened 2026-06-08
fix: ANR on init


tanchuev:
### Summary

`SoLoud.init()` performs the native AAudio device start **synchronously on the Dart calling thread**. When that
caller is the root isolate (the platform/UI thread on Android), a slow or contended audio HAL makes the call block
for >5 s, which trips Android's ANR watchdog. We are seeing this as a recurring production ANR on low/mid Android
devices.

### Environment

- flutter_soloud: 4.0.7 (latest)
- Platform: Android (AAudio backend), `in_foreground = true`, captured via `ApplicationExitInfo`
- Devices observed: Infinix X6525D (1.88 GB RAM), TECNO KL4 (Android 14, 3.97 GB RAM) — not limited to the very
  weakest hardware
- Call site: `await SoLoud.instance.init(bufferSize: 1024)` invoked once at app startup

### Impact

Recurring ANRs (multiple unique users) clustered on the first `init()` after cold start, especially when the audio
HAL is busy or when another AAudio operation (dispose/deinit on an app-lifecycle transition) runs concurrently.

### Root cause analysis

`FlutterSoLoudFfi.initEngine` is declared as a **synchronous** FFI function
(`ffi.NativeFunction<ffi.Int32 Function(ffi.Int, ffi.UnsignedInt, ffi.UnsignedInt, ffi.UnsignedInt)>`) and is
invoked synchronously (`final ret = _initEngine(...)`). The `await`s inside `SoLoud.init` only cover
`_initializeNativeCallbacks()` / `_loader.initialize()`, which run **after** `initEngine` has already returned — i.e.
**after** the native device start has already executed. Because Dart FFI calls run on the calling thread, the heavy
native work runs inline on the platform/UI thread.

Native call chain:

```
SoLoud.init
  -> FlutterSoLoudFfi.initEngine          (synchronous FFI, runs on the Dart calling thread)
    -> Player::init
      -> SoLoud::Soloud::init -> miniaudio_init
        -> ma_device_start__aaudio
```

Two blocking variants are observed in the wild:

1. **State-transition wait (most frequent / symbolicated):**
   `ma_device_start__aaudio -> ma_wait_for_simple_state_transition__aaudio -> AAudioStream_waitForStateChange -> syscall`.
   The stream has been requested to start; the calling thread then blocks waiting for the device to reach `STARTED`.

2. **Lock contention with a concurrent AAudio op:**
[...]

Issue #446, opened 2026-04-16
fix: small sound bleed after stop and play


Zerkley:
**Description**

When playing and stopping a looping sound. If I play another or the same sound, I can hear half a second of the old sound right where it stopped.

Video of the error:
[Video](https://youtu.be/yzyycclDem4)

I'm not sure if it's an error or is a bad soundHandle management on my part. I'll add the code that I have to load, play and stop.

**Steps to reproduce**

1. Load a few sounds
2. Play one with looping: true
3. Stop it
4. Play another sound or the same one
5. Previously stopped sound sounds again for .5 seconds

**Expected Behavior**

A stopped sound should not play it's last .5 seconds of audio when hitting play again.

**Additional Context**

```dart class MixerScreenController extends _$MixerScreenController {
  final Map<Object, SoundHandle> _activeBySoundKey = {};

  @override
  FutureOr<MixerScreenViewmodel> build() async {
    return await _getSounds();
  }

  Future<MixerScreenViewmodel> _getSounds() async {
    final soundRepository = ref.read(soundRepositoryProvider);
    try {
      final sounds = await soundRepository.getSounds();
      final soundsWithUrl = sounds
          .where((s) => (s.soundUrl ?? '').trim().isNotEmpty)
          .toList();
      final loadedSounds = await Future.wait(
        soundsWithUrl.map(
          (sound) =>
              soundPlayer.loadUrl(sound.soundUrl!, mode: LoadMode.memory),
        ),
      );
      return MixerScreenViewmodel(
        sounds: soundsWithUrl,
        loadedSounds: loadedSounds,
      );
    } catch (e) {
      rethrow;
    }
  }

  Object? _soundKey(Sound sound) {
    if (sound.soundId != null) return sound.soundId!;
    final url = sound.soundUrl;
    if (url != null && url.isNotEmpty) return url;
    return null;
  }

  AudioSource? _audioSourceFor(MixerScreenViewmodel vm, Sound sound) {
    for (var i = 0; i < vm.sounds.length; i++) {
      final s = vm.sounds[i];
      final same =
          s == sound ||
          (sound.soundId != null && s.soundId == sound.soundId) ||
          (sound.soundUrl != null &&
              sound.soundUrl!.isNotEmpty &&
              s.soundUrl == sound.soundUrl);
      if (same) return vm.loadedSounds[i];
    }
    return null;
  }

  Future<void> toggleSound(Sound sound, bool looping) async {
    final key = _soundKey(sound);
[...]

Issue #250, opened 2025-05-27
fix: Excessive battery drain due to AudioMix Wake Lock in Android


jahertor:
**Description**

I've observed excessive battery consumption in my Flutter application when using the flutter_soloud plugin in Android. An analysis of dumpsys batterystats shows a persistent AudioMix WakeLock that appears to be the culprit, even when the application is not actively playing audio or is in the background.

**Steps To Reproduce**

1. Configure audio_session in a Flutter application to manage an audio session.
2. Start an audio session.
3. Pause or stop the audio activity.
4. Leave the application in the background.
5. After some time, run `adb shell dumpsys batterystats --charged` to observe battery usage and wake locks.

I share you a gist with a sample code in order to :
https://gist.github.com/jahertor/9a40cfeeb8a1be358a20615d96a4db84

To reproduce, copy an audio file into assets/ folder and replace the first line of the code. Then add this dependencies on pubspec.yaml:
```
  flutter_soloud: ^3.1.8
  get: ^4.6.6
```

**Expected Behavior**

The AudioMix WakeLock should be released when audio playback is not active or has ended (`SoLoud.instance.stop(handle)` or at least when `SoLoud.instance.disposeAllSources()`) to minimize battery consumption. The background wake lock time should not be excessive if the application is not actively using audio.

**Screenshots**

![Image](https://github.com/user-attachments/assets/b5ca5ed4-0056-4027-a94e-b04a0ed14414)

**Additional Context**

Add the platform(s) where the issue arises or any other context about the problem here.
Tested on S20+ (Android 13) & S25 Ultra (Android 15)

Output of `adb shell dumpsys batterystats --charged`

```
  u0a329:
    Wake lock AudioMix: 3m 33s 98ms partial (0 times) max=222860 actual=222860 (running for 222860ms), 3m 42s 860ms background partial (0 times) max=222860 (running for 222860ms) realtime
    Wake lock *launch* realtime
    TOTAL wake: 3m 33s 98ms blamed partial, 3m 42s 860ms actual partial, 3m 42s 860ms actual background partial realtime
    Audio: 6m 45s 200ms realtime (0 times) (running)
    Foreground activities: 3s 516ms realtime (3 times)
    Screen discharge: 123uA for: 3s 516ms
    Total speaker time per level: 7:411
    Top for: 4s 406ms
    Cached for: 6m 41s 193ms
    Total running: 6m 45s 599ms
    Total cpu time: u=46s 60ms s=49s 985ms
[...]

PR #504, opened 2026-07-17
Add audio device control: stop/start/state APIs with isolate offloading

Colton127:
## Description

This PR adds new APIs for controlling miniaudio's audio output device:

- **`stopAudioDevice()`** / **`startAudioDevice()`**: Stop and restart the audio device while preserving loaded sounds, active voices, and engine state. Both are idempotent and run off the UI thread via isolates to prevent app freezing.
- **`getAudioDeviceState()`**: Query the current device state (uninitialized, stopped, started, starting, stopping). Synchronous and cheap since it's just an atomic read.
- **`deinitAsync()`**: Async variant of `deinit()` that runs the blocking native teardown off the UI thread, complementing the existing synchronous `deinit()` for contexts where awaiting is possible.
- **`setAndroidPauseDeviceWhenIdle()`**: Lets the Android implementation stop the audio device when idle, mirroring every other platform. Defaults to false to preserve the fix for https://github.com/alnitak/flutter_soloud/issues/446

Additionally, **`initEngine()` is now async** and runs off the UI thread via isolates, fixing ANR (Application Not Responding) issues on Android when device initialization takes several seconds (#481).

### Key Implementation Details

- **Isolate offloading**: Blocking native calls (`initEngine`, `startAudioDevice`, `stopAudioDevice`, `deinitAsync`) are wrapped in top-level functions (`_invokeInitEngine`, `_invokeDeviceLifecycle`, `_invokeVoidNative`) that can run in `Isolate.run()` workers. Only sendable primitives (ints) cross the isolate boundary; native function pointers are reconstructed inside the worker.
- **New enum**: `AudioDeviceState` mirrors miniaudio's `ma_device_state` for type-safe device state reporting.
- **C++ backend**: New miniaudio functions (`miniaudio_stopAudioDevice`, `miniaudio_startAudioDevice`, `miniaudio_getAudioDeviceState`) and Player methods expose device lifecycle control.
- **Web support**: Web backend stubs added (single-threaded, no isolates needed).
- **Shared teardown logic**: `_predeinit()` and `_postdeinit()` extracted to avoid duplication between `deinit()` and `deinitAsync()`.

## Type of Change

- [x] ❌ Breaking change (`initEngine()` signature changed from sync to async)

## Breaking Changes

- `initEngine()` now returns `Future<PlayerErrors>` instead of `PlayerErrors`. Callers must await the result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - fix: Waveform audio sources do not match engine sample rate #501. Thanks to @Colton127
 - add `setAndroidPauseDeviceWhenIdle` (opt-in, default off) to stop the audio device when idle on Android, releasing the audioserver `AudioMix` partial wakelock #250
 - add `stopAudioDevice()` / `startAudioDevice()` to stop and restart the audio output device without deinitializing the engine (loaded sounds and voices are preserved); the blocking native device operation runs off the UI thread so it does not freeze the app
+- add `getAudioDeviceState()` returning an `AudioDeviceState` enum to query the current state of the audio output device (uninitialized, stopped, started, starting, stopping)
 
 #### 4.0.12 (30 Jun 2026)
 - add `lowLatency` init option to allow recordable Android output #492. Thanks to @MjnMixael
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 #### 4.0.13 (XX Xxx 2026)
 - fix: Waveform audio sources do not match engine sample rate #501. Thanks to @Colton127
+- add `setAndroidPauseDeviceWhenIdle` (opt-in, default off) to stop the audio device when idle on Android, releasing the audioserver `AudioMix` partial wakelock #250
 
 #### 4.0.12 (30 Jun 2026)
 - add `lowLatency` init option to allow recordable Android output #492. Thanks to @MjnMixael
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 - add `setAndroidPauseDeviceWhenIdle` (opt-in, default off) to stop the audio device when idle on Android, releasing the audioserver `AudioMix` partial wakelock #250
 - add `stopAudioDevice()` / `startAudioDevice()` to stop and restart the audio output device without deinitializing the engine (loaded sounds and voices are preserved); the blocking native device operation runs off the UI thread so it does not freeze the app
 - add `getAudioDeviceState()` returning an `AudioDeviceState` enum to query the current state of the audio output device (uninitialized, stopped, started, starting, stopping)
+- fix: `init()` no longer blocks the UI thread — the blocking native engine/device initialization now runs off the UI thread, preventing ANRs on startup #481
+- add `deinitAsync()`, a non-blocking alternative to `deinit()` that runs the native teardown off the UI thread (the synchronous `deinit()` is unchanged)
 
 #### 4.0.12 (30 Jun 2026)
 - add `lowLatency` init option to allow recordable Android output #492. Thanks to @MjnMixael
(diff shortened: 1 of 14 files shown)


Colton127:
@alnitak This is still a work in progress. Running initEngine in an isolate is the right call and should have no tangible impact on performance/latency, it just prevents the UI thread from freezing. 

I encountered some trouble with the miniaudio's audio device lifecycle. Keeping the device active (outputting silence even when no sound is playing) is essential for low-latency sound playback. Otherwise, there is a measurable delay to begin playback.

However, keeping the audio device active even when idle introduces two major problems:

1. It maintains a partial wake lock on Android (https://github.com/alnitak/flutter_soloud/issues/250). Even after the app has been placed in the background. This is considered a "bad behavior" by Google and apps can be delisted as a result; https://android-developers.googleblog.com/2025/11/raising-bar-on-battery-performance.html. With this PR, a call to "stopAudioDevice" solves the issue while preserving all existing handles and audio sources.

2. iOS treats the audio device as an active sound. For those of us who use **audio_service**, this impacts play/pause playback controls in the control center (https://github.com/alnitak/flutter_soloud/pull/406). 

The fix to 2 solves the first, ~but introduces #446~ (edit: nvm, 446 is for web). 

My preferred solution is probably:
- Make the pause after 500ms of inactivity consistent across all platforms (it's only disabled on Android, and this PR addresses it with an option to re-enable it).
- Implement https://github.com/Colton127/flutter_soloud/pull/10 so developers can override the pause to keep the audio device alive while their app is active for low-latency playback.
- Add an optional parameter in soloud.init to allow deferring the audio device initialization until it's needed.

stopAudioDevice/startAudioDevice as well as getAudioDeviceState is added weight to the API but genuinely helpful. In fact it's a necessity for me as I need to init soloud and create audio sources and handles before playing anything.

alnitak:
@Colton127, thanks for your effort on this PR.

The changes look good to me. Also, your PR 10 looks good and I think it deserves to be included (here on in another PR).

My concern from the beginning has been to keep the engine as simple as possible for the user, but depending on the platform, that isn't always feasible. For the same reason, I'd also like to keep the [documentation](https://github.com/alnitak/flutter_soloud_docs/tree/main) as simple (and hopefully less scary 😄) as possible.

Also, just let me know when this PR is no longer a Work In Progress, and I'll be happy to give it a final review and merge it.

Colton127:
@alnitak I appreciate the prompt response!

https://github.com/Colton127/flutter_soloud/pull/10 is the way to go. It replaces the need for `setAndroidPauseDeviceWhenIdle` too, as it's the same function, just cross-platform.

How should the existing android lifecycle be handled?

1. Pause the audio device after 500ms of inactivity. This makes the behavior consistent across platforms (except web), but is a breaking change: it introduces a tangible delay in playback on Android when no sound was previously playing. This can be mediated via `setAudioDeviceKeepAlive(true)`.

2. Change the behavior to keep the audio device alive even when idle (`setAudioDeviceKeepAlive` defaults to `true`). Low-latency playback on all platforms.

3. Keep the existing behavior: Android keeps the audio device alive and maintains a partial wake lock. There's still a delay on iOS.

I'm learning toward 2 because it guarantees consistent, low latency across all platforms. The new API's in the PR allow full control over the lifecycle if needed.

alnitak:
@Colton127 thanks a lot for all this!

I am not very familiar with the audio behavior on mobile, so your effort is very welcome!

The "Pause the audio device after 500ms of inactivity" is a workaround I did without noticing the delay when starting the device again. So it's worth considering pausing the audio device after 500ms of inactivity only when the app is in the background? 

But I don't think the app lifecycle can be monitored from the plugin (in Dart), and maybe this should be made using method channels. Of course, your `setAudioDeviceKeepAlive` is still a valid functionality.

Colton127:
closed; replaced with https://github.com/alnitak/flutter_soloud/pull/508

PR #505, opened 2026-07-19
feat: add playClocked/play3dClocked with sample-accurate scheduling

### LINKED ISSUE #502: feat: Schedule playback at a precise instant in the future
**Description**

Hi there!

What I'm hoping for is something like Unity's [AudioSource.PlayScheduled(double time)](https://docs.unity3d.com/6000.0/Documentation/ScriptReference/AudioSource.PlayScheduled.html): a single call that takes an absolute (future) time on a persistent, queryable clock, and the engine handles playing samples at that moment precisely.

Main use case is a drift-free metronome. I found [issue #99](https://github.com/alnitak/flutter_soloud/issues/99) asked for basically this back in 2024, closed against PR #105. I read that example before filing this, but the solution seems to be, "use `Timer.periodic` and call `play()` on each tick." PR #105's buffer-size config reduces fixed latency (a real improvement), but it doesn't touch timer jitter, the actual source of drift I'm trying to avoid.

**Requirements**

- [ ] A way to schedule a sound to start at a precise, sample-accurate future time, not just "play these samples as close to right now as you possibly can".
- [ ] Scheduling needs to work far enough ahead to be useful for a metronome-like use case (seconds out, not just tens of milliseconds).
- [ ] A way to query the engine's current time from Dart, so a target time can be computed as an offset from "now."
- [ ] Works on iOS and Android.
- [ ] Clear documentation of the clock's units and reference point.

**Additional Context**

Also relates to [issue #498](https://github.com/alnitak/flutter_soloud/issues/498) (native loop-region end point). It's a different mechanism, but the same underlying theme of sample-accurate primitives that exist natively but aren't fully exposed to Dart yet.

Happy to test on real devices, or take a swing at a PR if you'd rather point me at an approach than write it yourself. 

Thanks for maintaining this, it's a great package!

alnitak:
## Description

Closes #502

Adds SoLoud's clocked-play APIs to the plugin, exposed on all platforms:

- `SoLoud.playClocked` / `SoLoud.play3dClocked` (+ `Bus.playClocked` / `Bus.play3dClocked`)
- `SoLoud.setDelaySamples` and `SoLoud.getStreamTime` (related scheduling APIs)

## Engine fix: clocked scheduling that actually works

Stock SoLoud `playClocked` resets its time reference every output buffer and clamps delays to a hardcoded 2048 samples, so with large buffers (or spacings > ~46 ms) sounds still snap to buffer boundaries — no better than `play()`.

Reworked the vendored engine (`soloud_core_basicops.cpp`, `soloud_core_3d.cpp`) with a persistent physics-time↔audio-sample anchor via `mStreamTime`.

Verified with a capture-based accuracy test:

| tick spacing | buffer 4096 | buffer 8192 | before fix (4096) |
|---|---|---|---|
| 50 ms | ±0.20 ms | ±0.20 ms | +43 ms (clumped) |
| 100 ms | ±0.35 ms | ±0.35 ms | up to +86 ms |
| 200 ms | ±0.49 ms | ±0.50 ms | up to +79 ms |

## Example & docs

- In the `metronome` example, there is now a "use playClocked" checkbox demonstrating an even rhythm, even with a 2048/4096/8192 buffer size.
- Pros/cons of `play` vs `playClocked` documented in the [docs](https://docs.page/alnitak/flutter_soloud_docs~dev/audio/playback#sample-accurate-scheduling-with-playclocked).

## Type of Change

- [x] 📝 Documentation


adammmckee:
Hey @alnitak, this is really nice work! For the metronome case, this looks great!

I have a second use-case that I think needs a couple more small additions. What would help me the most is:

1. A way to query the real current "now" used internally by soloud.
2. A single call to pin a future playback time as an offset from that real "now," handled atomically on the native side.
3. A way to schedule a played sound cutoff the same way: pinned to that real clock, not relative to when I make the call (the existing `scheduleStop` measures from call time, which doesn't work for a sound whose start is itself scheduled seconds in advance).

Concretely, something like:

```dart
double getStreamTime();

SoundHandle playScheduled(
  AudioSource source,
  double atTime, {
  double volume = 1,
  double pan = 0,
});

void stopScheduled(SoundHandle handle, double atTime);
```

That combination would let me build a "playback manifest" of scheduled notes at a tempo (think music score playback) against the real clock, including each note's duration, and would let me cancel anything still pending if the user pauses playback. Something like:

```dart
// Query the real clock once, schedule a batch of upcoming notes against it.
final now = SoLoud.instance.getStreamTime();
final manifest = <({SoundHandle handle, double atTime})>[];
for (final note in upcomingNotes) {
  final atTime = now + note.offsetFromNow;
  final handle = SoLoud.instance.playScheduled(note.audioSource, atTime);
  SoLoud.instance.stopScheduled(handle, atTime + note.duration);
  manifest.add((handle: handle, atTime: atTime));
}

// User hits pause: cancel anything scheduled but not yet audible.
final cutoff = SoLoud.instance.getStreamTime();
for (final entry in manifest) {
  if (entry.atTime > cutoff) {
    SoLoud.instance.stop(entry.handle);
  }
}
```

As it stands, there are a few things that playClocked (and the existing scheduleStop) don't quite support in this draft PR:

1. Calls landing more than ~2 seconds out get treated as a clock jump and re-anchored, so I can't pre-schedule a wider window at once.
2. Composing my own version from the exposed setDelaySamples has a gap: the delay only counts down once a voice is unpaused, so the round-trips between reading "now" and actually unpausing would add a small, uncompensated delay that playClocked's own single native call avoids.
3. scheduleStop schedules relative to call time, which doesn't compose with notes whose start is itself scheduled in the future.

Would you be open to adding something like playScheduled/stopScheduled plus a global getStreamTime()? Happy to test on real devices for you and/or make a PR myself in this repo - let me know what would help you the most!

Thanks again! You're helping a ton of people with soloud (myself included)!

alnitak:
@adammmckee I have split from here the new PR #509 with your suggestions (with some more features! :) )

PR #509, opened 2026-07-25
feat: add `getEngineTime`, `playScheduled`, `stopScheduled` and `fadeScheduled` (extends #505)

alnitak:
## Description

Follow-up to #505, addressing @adammmckee comment.

While `playClocked` schedules sounds against the caller's own clock (within a ~2 s window), this PR pins scheduling to the engine's own stream clock, exposed as `SoLoud.instance.getEngineTime()`:
- **`playScheduled(sound, atTime, {duration, busId, volume, pan})`** — sample-accurate start at an absolute engine time, computed atomically on the native side. No anchor and no re-anchor guard, so sounds can be scheduled arbitrarily far in the future (the >2 s limitation of `playClocked` does not apply). The optional `duration` stops the sound at `atTime + duration` in the same call.
- **`stopScheduled(handle, atTime)`** — stop at an absolute engine time, composing with sounds whose start is itself scheduled (unlike `scheduleStop`, which measures from call time).
- **`fadeScheduled(handle, atTime, to, time, {thenStop})`** — volume fade starting at an absolute engine time, with optional stop at fade end.
- **`Bus.playScheduled`** mirrors `Bus.playClocked`.

Scheduled stops are sample-accurate (a new per-voice sample countdown in the mixing loop, mirroring `mDelaySamples`), not quantized to output-buffer boundaries like `scheduleStop`. The relative `scheduleStop` is unchanged.

Added `playClocked` and `playScheduled` tests.

This enables score/manifest-style playback: read the clock once, schedule a batch of notes with their durations, and cancel anything still pending on pause.

Also the metronome now lets you compare `play` / `playClocked` / `playScheduled`; new `scheduled_accuracy_test` verifies sample-accurate starts (0.00 ms deviation), far-future scheduling, and sub-buffer cutoffs.

New docs [here](https://docs.page/alnitak/flutter_soloud_docs~dev/audio/playback#scheduling-on-the-engine-clock-playscheduled).

@adammmckee feel free to test on your devices! I still need some more tests on mines.

## Type of Change

- [X] 📝 Documentation


adammmckee:
Hey @alnitak!

This looks great! I'll give this a try today/tomorrow and provide some feedback from testing on a real device.

Thanks again!

adammmckee:
Good timing on these updates, @alnitak - I'm building a test harness in my app for your PR right now. I've pulled them down and I'll have some feedback for you today!

adammmckee:
Hey @alnitak, I found a small issue (and proposed a solution) with this branch while running on an iOS physical device:

`src/flutter_soloud.cpp` is missing an include that `src/CMakeLists.txt` has:

```
#include "mixeroutput/wav_output_encoder.cpp"
```

It should be added after the flac_output_encoder.cpp line (around line 129). Without it, SwiftPM builds crash at launch with `dyld: symbol not found in flat namespace '__ZTV16WavOutputEncoder'`.

I added this include locally, and my app built.

Hope this helps!

adammmckee:
Hey again, I found another runtime crash on iOS, and I found what seems to be the solution.

`src/soloud/src/core/soloud.cpp`, in `mix_internal` (around line 2235):

The new "Scheduled absolute-time stop" block dereferences mVoice[i] without checking it's non-null first.

The mStopScheduler block directly above it can call stopVoice_internal(i), which sets mVoice[i] = 0 and deletes the instance. So if that fires, the very next block crashes on a null/freed pointer.

Since scheduleStop's duck-then-stop path hits this on essentially every note release, it's not scheduling-specific — it reproduces on ordinary play/noteOff too.

Fix: guard the new block the same way stopVoice_internal guards its own access:

```
if (mVoice[i] && mVoice[i]->mStopSamplesLeft >= 0 && ...)
```

I applied this guard locally, and it resolves the crash.

Hope this helps!

alnitak:
Great @adammmckee! I forgot the SPM build for tests, thank you. I tested only using CocoaPods.

And since you are replying quickly, I should tell you that I have to fix this tomorrow :)

adammmckee:
Awesome, thanks so much @alnitak!

I validated playScheduled/getEngineTime end-to-end for score playback works on real iOS and Android devices.

Here's what I validated:

I built a small score-playback feature specifically to exercise this API for a real use case: read the engine clock once via getEngineTime(), then schedule an entire score's worth of notes against that single anchor via playScheduled(..., duration:), rather than triggering notes one at a time as they come due.

What I tested:

1. Created a score in a notation editor, saved it, then reopened it from a separate playback screen
2. Played it back through the scheduler — a real piece, not a synthetic test case, about 10 seconds long with a mix of note durations
3. Changed tempo and replayed the same score at the new tempo (re-deriving the schedule and re-anchoring against a fresh getEngineTime() read each time)
4. Ran all of the above on iOS and Android

Result: it does what I need. The whole batch-of-notes-against-one-clock-read model is exactly the workflow I was hoping for, and once the two bugs below were fixed, playback was clean and in time on both platforms.

Two bugs found and fixed locally while testing (already posted separately, noting here for a complete picture):

1. SwiftPM builds crash at launch (dyld: symbol not found in flat namespace '__ZTV16WavOutputEncoder') — src/flutter_soloud.cpp's unity-build include list is missing mixeroutput/wav_output_encoder.cpp, present in src/CMakeLists.txt but not here.
2. Null/use-after-free in mix_internal (soloud.cpp ~line 2235) — the new scheduled-stop block dereferences mVoice[i] without checking it wasn't just nulled/freed by the mStopScheduler block directly above it. Hits on ordinary noteOff/scheduleStop, not just the new scheduling calls.

Looking forward to applying your fixes tomorrow and trying making sure it all works together!

alnitak:
@adammmckee I have committed your fixes.

I still have some problems with an old Android device (I don't know yet if it is related to this PR), but if it's fine at your end, once I've solved it, I'll be able to merge.

adammmckee:
Hey @alnitak!

Ran into an undocumented limitation while testing score playback and thought I'd bring it up.

soloud.dart's doc for `setMaxActiveVoiceCount` says "The hard maximum count is 4095," but `mapResampleBuffers_internal` (soloud.cpp:1956) asserts `mMaxActiveVoices < 256`, backed by a fixed-size stack buffer:

```
SOLOUD_ASSERT(mMaxActiveVoices < 256);
char live[256];
memset(live, 0, mMaxActiveVoices);
```

Is it possible to rework this assertion for 255 max voices to allow me to allocate up to 4095? Or is there another reason for asserting that 255 is the max number of voices?

Thanks again - the changes in this PR exactly solve my problem!

alnitak:
> Thanks again - the changes in this PR exactly solve my problem!

Great!

Regarding the max active voice count, you caught me unprepared, I must dig around the SoLoud lib. I think that we should look at the [VOICE_COUNT](https://github.com/alnitak/flutter_soloud/blob/d3c9f7c8c5eb98507557d34ef4a32e92924cb8c8/src/soloud/include/soloud.h#L89) definition because this has been checked in [soloud_core_setters.cpp](https://github.com/alnitak/flutter_soloud/blob/d3c9f7c8c5eb98507557d34ef4a32e92924cb8c8/src/soloud/src/core/soloud_core_setters.cpp#L75). For sure, the `setMaxActiveVoiceCount` doc in `soloud.dart` should be updated referencing the 1024 limit and not telling the user about the 4095 hard limit (which is internal to the lib).

I don't know yet how `mMaxActiveVoices` and `VOICE_COUNT` are related. That assert sounds odd. I'll be back to you if I find something.

alnitak:
@adammmckee I really think that was a latent bug. The assert only works in debug mode, so if in production the `mMaxActiveVoices` is greater than 256, a stack overflow is thrown.

I left the `VOICE_COUNT` untouched at 1024. Do you really need more?

alnitak:
@Kunstderfug I forgot to mention you. I think this new feature (not the VOICE_COUNT issue) might be of interest to you for your work.

Kunstderfug:
@alnitak Thank you! Yaaas! My app is based on scheduling so this new level of precision would be awesome to try.

adammmckee:
Hey @alnitak!

Sorry, I was on the road all day and just saw your comment. 1024 voices is more than enough for any practical case I can think of. Thanks so much! Looking forward to this making its way to pub.dev!

alnitak:
@adammmckee version 4.1.0 is now available on pub.dev!

Thanks for all

adammmckee:
@alnitak Amazing! Thank you so much! Looking forward to collaborating again next time!

Kunstderfug:
Awesome!! ✌️✌️✌️

adammmckee:
Hey again @alnitak!

I think I found another one-liner fix for SPM builds for iOS in 4.1.0 😄 

I updated my project to pull 4.1.0 from pub.dev (instead of my local package clone), and I get this error when building for iOS via SPM:

```
Could not build the precompiled application for the device.
Semantic Issue (Xcode): Redefinition of 'MixerOutputEncoder'
flutter_soloud-4.1.0/src/mixeroutput/mixer_output_encoder.h:11:6

Semantic Issue (Xcode): Redefinition of 'PcmChunkQueue'
flutter_soloud-4.1.0/src/mixeroutput/pcm_queue.h:11:6

Semantic Issue (Xcode): Redefinition of 'MixerOutput'
flutter_soloud-4.1.0/src/mixeroutput/mixer_output.h:22:6
```

The cause seems to be that `src/soloud/src/backend/miniaudio/soloud_miniaudio.cpp:73` includes a mixeroutput header without a relative prefix:

```
#include "mixeroutput/mixer_output.h"
```

Looks like the fix is to update the path on the include to be relative to itself:

```
#include "../../../../mixeroutput/mixer_output.h"
```

I tested this against the actual pub.dev package (not just my local clone, which hard-links the two `src/` copies and would build either way). I confirmed it fails without this change and builds clean with it, from a fully clean DerivedData both times.

Any chance we could get this released in 4.1.1? 😄

alnitak:
That's strange because the [`Package.swift`](https://github.com/alnitak/flutter_soloud/blob/7b6a07ff60f916071f84f362da4d83742e92bc58/ios/flutter_soloud/Package.swift#L48) has the headers search path for the `metronome` dir!

Anyway, adding the relative path is not a problem, and I think I have to do this.

Version 4.1.1 published with the fix 😄

adammmckee:
Outstanding! I'm fully online with 4.1.1 with the pub.dev reference in my workflow on iOS and Android. All builds are green!

Once again, thank you so much @alnitak!

PR #518, opened 2026-07-29
Fix voice-ended callback deadlock

Colton127:
## Description

Fixes a deadlock that occurs when a voice ends while another sound is being disposed. The issue stems from a lock-order inversion:

- `stopVoice_internal()` runs with SoLoud's audio mutex held and directly invoked the embedder's voice-ended callback
- The callback reaches back into `Player` state guarded by `sounds_mutex`
- Meanwhile, `Player::disposeSound()` holds `sounds_mutex` and calls `soloud.stop()`, which needs the audio mutex
- Result: deadlock with the audio mutex stranded, blocking all subsequent SoLoud calls including `deinit()`

**Solution:** Queue ended voice handles and dispatch callbacks *after* releasing the audio mutex, eliminating the lock-order inversion. The dispatch loop handles re-entrant SoLoud calls from callbacks (which re-acquire the mutex) and prevents recursion by guarding against nested dispatches.

## Changes

- **soloud.h**: Made `_voiceEndedCallback` atomic for thread-safe reads during dispatch; added queue and guard fields (`mEndedVoiceQueue`, `mEndedVoiceCount`, `mDispatchingEndedVoices`)
- **soloud_core_voiceops.cpp**: Changed `stopVoice_internal()` to queue handles instead of calling the callback directly
- **soloud.cpp**: Added `unlockAudioMutexAndDispatchEndedVoices_internal()` to drain the queue after releasing the mutex; modified `unlockAudioMutex_internal()` to trigger dispatch when needed
- **CHANGELOG.md**: Documented the fix
- **test/voice_ended_callback_test.cpp**: Comprehensive regression test suite covering callback dispatch, re-entrancy, lock-order inversion, and edge cases
- **test/run_voice_ended_callback_test.sh**: Build script for the test suite

## Type of Change

- [x] ✅ Build configuration change (test infrastructure)

## Test Plan

Added 8 regression tests in `voice_ended_callback_test.cpp` that verify:
- Callbacks run with the audio mutex released
- Natural end-of-stream dispatch works correctly
- Multiple ended voices arrive in queue order
- Callbacks can safely re-enter SoLoud
- Re-entrant voice stops maintain FIFO order without recursion
- Unregistering the callback mid-batch suppresses remaining handles
- Null callbacks drain the queue without dispatching
- The original lock-order inversion scenario completes without deadlock

Tests include watchdog timers to detect deadlocks and fail fast rather than hang. Run with `./test/run_voice_ended_callback_test.sh`.


alnitak:
Hi @Colton127, thanks for this PR.

I'm always a little afraid of touching the core engine and adding complications to the code that will make it difficult to fix any future bugs.

This fix seems necessary, and I thank you!

PR #532, opened 2026-08-04
fix: changeDevice() device selection and error reporting

Colton127:
## Description

Fixes two correctness defects in `SoLoud.changeDevice()`, plus three related ones found in the same code while making those correct.

### `changeDevice()` could never select the default device

`SoLoud.instance.changeDevice()` with no argument is documented to select the system default output device, but it always threw `SoLoudNoPlaybackDevicesFoundCppException`. Dart passes `-1`, and `Player::changeDevice()` compared that `int` against `devices.size()`:

```cpp
if (devices.size() == 0 || deviceID >= devices.size())
    return noPlaybackDevicesFound;
```

`-1` is promoted to `SIZE_MAX` by the usual arithmetic conversions, so the guard always fired. The device list is now consulted only for explicit (non-negative) IDs, and `-1` passes `nullptr` straight through to `deviceConfig.playback.pDeviceID`, which is miniaudio's default-device selector. This mirrors what `Player::init()` already did. IDs below `-1` are rejected as `invalidParameter` before any conversion to an unsigned type.

### A failed device change reported success

```cpp
if (result != SoLoud::SO_NO_ERROR)
    result = backendNotInited;   // assigns to a local
return noError;                  // ...then returns success regardless
```

If miniaudio could not initialize or start the replacement device, the caller was told the switch worked. It now returns `audioDeviceFailedToStart`, which Dart already maps to `SoLoudAudioDeviceFailedToStartCppException`. `backendNotInited` would be misleading here: the engine can stay initialized while its output device could not be replaced.

Note this PR only makes the failure *visible* — it deliberately does not add rollback, retry, or fallback-to-default. 

### Also fixed in the same code

- **Use-after-free in `Player::init()`.** With an explicit device ID, the device list was built inside an `if` block and `playbackInfos_id` pointed at one of its elements. The vector was destroyed at the end of that block, so `soloud.init()` handed miniaudio a pointer into freed heap. Reproduced under AddressSanitizer (details below).
- **`init()` had the same signed/unsigned comparison**, reporting IDs below `-1` as `noPlaybackDevicesFound`. Now `invalidParameter`, matching `changeDevice()`.
- **`listPlaybackDevices()` leaked on every call.** `PlaybackDevice::name` was a `strdup()`ed `char *` that nothing ever freed — one allocation per device per call, including the internal calls from `init()` and `changeDevice()`. It is now a `std::string`, which also lets the dangling `pPlaybackInfos` member become a local: it points at context-owned memory that `ma_context_uninit()` frees before the function returns.
- **Out-of-bounds read in the device-name filter.** `bindings.cpp` scanned the first 5 bytes of each name for control characters *before* checking `strlen(name) <= 5`, over-reading the buffer for short names. The length check now runs first; filtering behaviour is unchanged.
- **`Soloud::miniaudio_changeDevice()` returned 0 when the audio thread mutex was null**, reporting a successful device change when nothing had happened. It now returns `UNKNOWN_ERROR` (surfacing as `audioDeviceFailedToStart`), and `NOT_IMPLEMENTED` when built without the miniaudio backend.

### Verification

The use-after-free and the leak were confirmed with a standalone AddressSanitizer harness built from the real plugin sources (miniaudio's null backend enumerates a device, so the explicit-device path is reachable without audio hardware).

Against the pre-fix code:

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 256 at 0x5120000001e8
  #1 ma_device_init            miniaudio.h:43662
  #4 SoLoud::miniaudio_init    soloud_miniaudio.cpp:540
  #5 SoLoud::Soloud::init      soloud.cpp:335
  #6 Player::init              src/player.cpp:273
freed by thread T0 here:
  #5 std::vector<PlaybackDevice>::~vector()
  #6 Player::init              src/player.cpp:269
```

...and `132 byte(s) leaked in 2 allocation(s)`, both `strdup` from `Player::listPlaybackDevices()`. The patched build is silent on both. Worth stressing that a passing run proves little on its own here — freed storage often still holds the right bytes, so the buggy version could return `noError` and play normally.

### Tests

`example/tests/tests/playback_devices.dart` now covers the default selector, an enumerated device, rejected native selectors (`-2` → `invalidParameter`, out-of-range → `noPlaybackDevicesFound`), and initializing on an explicit device — the use-after-free path, which nothing exercised before.

### Two things to note

1. **The WASM artifact has not been rebuilt.** `src/player.cpp` is compiled into `web/libflutter_soloud_plugin.wasm`, so Web keeps the old behaviour until it is regenerated. I hit an unrelated failure building the Xiph libs locally (`vorbis` configure: `must have Ogg installed!` — `--with-ogg="$OGG_DIR"` expects an installed prefix layout, but the script builds ogg in-tree so the archive is at `$OGG_DIR/src/.libs/`). Since the artifact is normally yours to commit anyway, could you regenerate `web/libflutter_soloud_plugin.{js,wasm}` on merge? No exported signatures changed.
  
2. The changelog entry is filed under a placeholder `##### 4.1.7 (X Xxx 2026)` heading — retitle or fold it in as you see fit. No version bump included.
  

No ABI, enum, exception, or binding signature changes.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


alnitak:
Hi @Colton127, again, thank you very much for your support!

> Since the artifact is normally yours to commit anyway, could you regenerate web/libflutter_soloud_plugin.{js,wasm} on merge?

Sure! I had no problems building wasm. For the record, I am using em++ version 4.0.22-git on macOS, but maybe this is not the problem.

Thanks again. Merging

PR #533, opened 2026-08-06
fix: run engine init and teardown off the UI isolate

### LINKED ISSUE #481: fix: ANR on init

Colton127:
## Summary

Moves blocking native engine initialization and teardown off Flutter’s root/UI isolate.

Fixes #481.

## Problem

`SoLoud.init()` invoked native `initEngine()` through synchronous Dart FFI.

On Android, engine initialization starts miniaudio’s AAudio device. Miniaudio may synchronously wait for the stream to reach its started state or block while another AAudio operation holds an internal lock.

When `init()` is called from Flutter’s root isolate, this native work runs on Android’s platform/UI thread. A slow or contended audio HAL can therefore freeze rendering and input long enough to produce an ANR.

Delaying the call, leaving its future unawaited, or wrapping it in a Dart timeout does not move the synchronous native work away from the UI thread.

## Solution

The native `initEngine()` call now runs through `Isolate.run()` on native platforms.

The worker receives only the native function address and primitive initialization arguments. It invokes the blocking native function and returns its result.

Dart callbacks are not created inside the worker. `NativeCallable` creation, callback registration, loader initialization, and Dart readiness publication still occur on the original Flutter isolate after native initialization completes.

Web retains a direct implementation behind the asynchronous binding interface.

## Lifecycle ordering

Moving init and teardown onto workers allows native execution order to differ from Dart request order. The change therefore includes the lifecycle coordination needed to preserve correct behavior:

* Initialization requests are serialized through a FIFO future chain.
* Every initialization captures a lifecycle generation when requested.
* A later deinit invalidates active and queued initialization requests.
* Native shutdown is published synchronously before async teardown is dispatched.
* Native initialization checks for shutdown before starting and before publishing readiness.
* Native readiness is exposed through a lock-free atomic rather than the replaceable player pointer.
* Internal initialized state is atomic.
* Concurrent async teardown calls share one native teardown.
* Dart callbacks remain alive until native teardown has completed.
* Stale-engine cleanup after hot restart uses async teardown without cancelling the initialization that requested the cleanup.

When shutdown interrupts initialization, the init future completes with `SoLoudInitializationStoppedByDeinitException`.

## Test adjustment

The existing `AsynchronousDeinit` example test launched initialization with `unawaited()` and did not wait for its completion handler.

With initialization now genuinely asynchronous, those handlers could outlive the test and call `deinit()` during the following test.

The test now:

* retains the initialization future;
* awaits `deinitAsync()`;
* awaits the initialization result;
* accepts the intentional initialization-cancelled exception;
* rethrows unexpected errors with their original stack;
* verifies the native engine is uninitialized before continuing.

This prevents asynchronous lifecycle work from leaking between test iterations or into subsequent tests.

## Scope

This change does not remove or shorten miniaudio’s AAudio state-transition wait.

A slow audio HAL may still delay or fail native initialization, but the wait now occurs on a worker rather than freezing Flutter’s UI thread.

This PR does not include broader device lifecycle APIs, device schedulers, interruption recovery, lazy startup, manual start/stop controls, FlutterEngine detach handling, or unrelated audio changes.

## Compatibility

The internal `initEngine()` binding changes from synchronous to asynchronous, but it is not part of the public package API.

Public `SoLoud.init()` remains asynchronous.

The existing synchronous `deinit()` method remains available, and `deinitAsync()` provides nonblocking teardown where required.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


alnitak:
Hi @Colton127, thanks again for your PR!!

I ran the tests and found that on Android the `PlaybackDevices` test fails with this:
```
D/AAudio  (26126): AAudioStream_requestStart(s#10) returned 0 ---------
D/AudioStreamInternalPlay_Client(26126): callbackLoop() entering >>>>>>>>>>>>>>>
V/flutter_soloud NDK(26126): miniaudio_changeDevice_impl: ma_device_start failed with error -1
```

I think this doesn't depend on this PR, but maybe on #532. Maybe it is related to the changes in the `miniaudio_changeDevice_impl` function of `soloud_miniaudio.cpp`.

I just tried using v4.1.6 and it wasn't happening. If you want, or if you do not have the time, I can take a look tomorrow.

Colton127:
> Hi @Colton127, thanks again for your PR!!
> 
> I ran the tests and found that on Android the `PlaybackDevices` test fails with this:
> 
> ```
> D/AAudio  (26126): AAudioStream_requestStart(s#10) returned 0 ---------
> D/AudioStreamInternalPlay_Client(26126): callbackLoop() entering >>>>>>>>>>>>>>>
> V/flutter_soloud NDK(26126): miniaudio_changeDevice_impl: ma_device_start failed with error -1
> ```
> 
> I think this doesn't depend on this PR, but maybe on #532. Maybe it is related to the changes in the `miniaudio_changeDevice_impl` function of `soloud_miniaudio.cpp`.
> 
> I just tried using v4.1.6 and it wasn't happening. If you want, or if you do not have the time, I can take a look tomorrow.

I'm looking into it. I can confirm it surfaced in [#532](https://github.com/alnitak/flutter_soloud/pull/532). The test passes on my primary devices - iPhone 11 and S26 Ultra. However, I encountered the exact "ma_device_start failed with error -1" on an Android emulator and LG V20. 

The issue occurs at ```SoLoud.instance.changeDevice``` and does so regardless of if a device is passed or not.

I'm going to first surface the exception. The test app shouldn't freeze; if it encounters a problem, it should be thrown correctly. 

Edit: Resolved in https://github.com/Colton127/flutter_soloud/pull/27. Merged into this PR. I added regression tests and ensured exceptions are thrown correctly instead of blocking the UI thread. Tests passed on all my devices. WASM still needs to be rebuilt.

alnitak:
Great! Thank you very much!

Merging, and maybe later today I'll publish on pub.dev

PR #535, opened 2026-08-09
fix: Tie native engine lifecycle to FlutterEngine

### LINKED ISSUE #126: fix: handles seemingly become invalid after leaving standby (on Android)
**Description**

So we have a few Pixel phones, that, when the devices have been in standby for a while, stop playing cached sounds after waking up again.

Maybe we're doing it wrong.

We basically cache them like this:

```dart
AudioSource? click;
AudioSource? addItem;
AudioSource? delete;
AudioSource? purchase;
AudioSource? scan;

Future<AudioSource> getAudioSource(SoLoud instance, String soundName) async {
  const root = 'assets/sounds';
  if (soundName == 'click.wav') {
    click ??= await instance.loadAsset('$root/click.wav');
    return click!;
  } else if (soundName == 'add-item.wav') {
    addItem ??= await instance.loadAsset('$root/add-item.wav');
    return addItem!;
  } else if (soundName == 'delete.wav') {
    delete ??= await instance.loadAsset('$root/delete.wav');
    return delete!;
  } else if (soundName == 'purchase.wav') {
    purchase ??= await instance.loadAsset('$root/purchase.wav');
    return purchase!;
  } else if (soundName == 'scan.wav') {
    scan ??= await instance.loadAsset('$root/scan.wav');
    return scan!;
  } else {
    throw Exception('Unknown sound filename');
  }
}
```

And then the "cached" AudioSources are using like so:

```dart
Future<void> feedback([String? sound]) async {
  final sl = SoLoud.instance;
  if (!await sl.initialized) await sl.init();
  final source = await getAudioSource(sl, sound ?? 'click.wav');
  final handle = await sl.play(source);
  // Attempted fix of handles dying, but it doesn't work.
  if (!sl.getIsValidVoiceHandle(handle)) {
    debugPrint('SoundHandle error');
    click = null;
    addItem = null;
    delete = null;
    purchase = null;
    scan = null;
    await sl.disposeAllSources();
  }
}
```

We instantiate SoLoud in main.dart like so:
```dart
final sl = SoLoud.instance;
if (!await sl.initialized) await sl.init();
```

**Steps To Reproduce**

1. Use the above code
2. Verify that sounds play
3. Minimize the app in Android
4. Turn the device off for a while
5. Maximize the app again, and try to play the sound. Nothing should play at this point until the app is restarted.

**Expected Behavior**

I expect that assets stay loaded, even after leaving standby.

Colton127:
Fixes https://github.com/alnitak/flutter_soloud/issues/126

## Description

Implements FlutterEngine lifecycle management for the process-global native audio engine. The native engine (SoLoud player, output device, lifecycle scheduler, and Dart callback pointers) now stays synchronized with the FlutterEngine that owns it, preventing callbacks from being invoked after their isolate is destroyed.

### Key Changes

**Core Lifecycle Management** (`src/bindings.cpp`, `src/engine_lifecycle.h`):
- Added `clearDartCallbackRegistrationsForEngine()` to retire all Dart callables when an engine detaches or hot-restarts
- Added `requestEngineTeardownForEngine()` to tear down the native engine when a FlutterEngine is destroyed
- Implemented generation-based callback invalidation that makes every callable inert synchronously without walking the sound list
- Added `prepareEngineInitForRequest()` to handle asynchronous engine claims with epoch-based validation

**Callback Gate** (`src/dart_callback_gate.h`):
- New synchronization primitive that gates all Dart callable invocations through a shared lock
- Uses generation numbers to invalidate callbacks without reaching per-source callables
- Ensures no invocation is in flight when retirement returns and prevents new invocations afterward

**Platform Plugins**:
- **Android** (`android/src/main/java/flutter/soloud/flutter_soloud/FlutterSoloudPlugin.java`): Observes FlutterEngine lifecycle via `FlutterPlugin` interface, calls native lifecycle hooks on attach/detach
- **iOS** (`ios/flutter_soloud/Sources/flutter_soloud/FlutterSoloudPlugin.mm`): Receives engine ID over method channel, triggers teardown in `detachFromEngineForRegistrar:`
- **macOS** (`macos/flutter_soloud/Sources/flutter_soloud/FlutterSoloudPlugin.mm`): Similar to iOS, receives engine ID and manages lifecycle

**Dart Integration** (`lib/src/bindings/`):
- Added `DarwinEngineLifecycle` for iOS/macOS to hand engine ID to native code via method channel
- Updated `FlutterSoLoudFfi` to read `PlatformDispatcher.instance.engineId` and pass it to native init
- Added async prepare path for platforms requiring asynchronous engine claims

**Audio Streams** (`src/audiobuffer/`):
- `BufferStream` and `PullBufferStream` now record the callback generation when registered
- Stream callbacks check generation before invocation to stay inert after retirement

### Behavior

- **Hot Restart**: Immediately retires all callbacks (global and per-stream) so the audio thread cannot call into the discarded isolate
- **Engine Destroy**: Tears down the native engine and stops the output device
- **Replacement Engine**: New registrations do not resurrect retired sources; old sources stay dead even under new callbacks
- **Concurrent Operations**: Both lifecycle hooks return promptly while device operations are in progress, preventing ANRs on Android

### Testing

Added comprehensive native regression test suite (`test/engine_lifecycle_test.cpp`) covering:
- Callback retirement for hot restart
- Replacement engine isolation
- Callback ownership vs. lifecycle ownership separation
- Concurrent teardown scenarios
- Duplicate destroy/detach handling
- Platform thread responsiveness

Build and run with: `./test/run_engine_lifecycle_test.sh`

## Type of Change

- [x] 🧹 Code refactor


alnitak:
Hi @Colton127, this is a huge PR, and it took me some time to thoroughly test it. Everything looks good, and I really hope this will finally fix the old #126!

After merging, I’ll move the `Output-Device Lifecycle` section from README.md to [flutter_soloud_doc](https://github.com/alnitak/flutter_soloud_docs). I’ll also restore the `mixer output capture` and `mixing bus` features to the Features section, as I think they are important features worth highlighting.

Thank you again for your work on this PR.

PR #541, opened 2026-08-28
build system migration to Dart build hooks

alnitak:
## Description

- **breaking change: build system migration to Dart build hooks** (https://dart.dev/tools/hooks):
  - Android, iOS, macOS, Linux and Windows native code is now compiled by `hook/build.dart` (`package:hooks` + `package:native_toolchain_c`) instead of per-platform plumbing (Android CMake, iOS/macOS CocoaPods script phases, SwiftPM unity build). No CMake or podspec script phases are needed anymore.
  - The Xiph libraries (Opus/Ogg/Vorbis/FLAC) are linked (statically on Apple, bundled as shared libraries on Android/Windows).
  - `NO_XIPH_LIBS` is now configured via hook user-defines in the *app* pubspec instead of an environment variable:
    ```yaml
    hooks:
      user_defines:
        flutter_soloud:
          no_xiph_libs: true
    ```
    [It is not possible to expose an environment variable to Hooks](https://dart.dev/tools/hooks#environment-variables). But some env variables are passed to the Hooks sandbox `NIX_*` included, and maybe a trick could be to use something like `NIX_NO_XIPH_LIBS` even if those env variables are coming from NixOS? /cc @filiph 
  - The FFI surface in `src/bindings.cpp` now has a real header, `src/bindings.h`, which is the single ffigen entry point (the hand-maintained `src/ffi_gen_tmp.h` is gone).
  - Dart bindings now use `@Native` with asset id `package:flutter_soloud/src/bindings.cpp`; the native library is resolved by the VM, so no `DynamicLibrary.open`/force-load/`STRIP_STYLE` workarounds are needed.
  - **Web is unchanged**: hooks do not cover web, the emscripten build in `web/` is still used.

## Type of Change

- [X] 🧹 Code refactor
- [X] 📝 Documentation


filiph:
Fantastic, looking forward to this!

> [It is not possible to expose an environment variable to Hooks](https://dart.dev/tools/hooks#environment-variables). But some env variables are passed to the Hooks sandbox NIX_* included, and maybe a trick could be to use something like NIX_NO_XIPH_LIBS even if those env variables are coming from NixOS? /cc @filiph

Do I understand correctly that:

1. Disabling Xiph libs is still possible, but now it's through a yaml file? (That's actually better than an env variable, in my book.)
2. There's no way to tell the developers that `NO_XIPH_LIBS` no longer works -- because the tool that read `NO_XIPH_LIBS` in the past is being removed?
3. So people who want to get rid of xiph libraries need to be reached somehow (CHANGELOG, social media) to learn about this big build system change.
4. The failure state that I was worried about the most (that the app is built correctly but then fails at runtime because xiph libs aren't installed on the user's system) is no longer possible? Xiph libs, if used, are _always_ shipped with the app?

alnitak:
Perfect understanding!

> 1. Disabling Xiph libs is still possible, but now it's through a yaml file? (That's actually better than an env variable, in my book.)

Yes, no more the env variable, only through `pubspec.yaml`.

> 2. There's no way to tell the developers that `NO_XIPH_LIBS` no longer works -- because the tool that read `NO_XIPH_LIBS` in the past is being removed?

Exactly, the Hooks build is sandboxed, and AFAIK, the native build outputs (stderr and stdout) are inside the `.dart_tools/hooks_runner/flutter_soloud` and no logs are available to the user while building the native.

> 3. So people who want to get rid of xiph libraries need to be reached somehow (CHANGELOG, social media) to learn about this big build system change.

Hope that changing to a major version will reduce the number of users who are not curious to look at the changelog :) (have you recently seen how maaany weekly downloads we have? Impressive).

> 4. The failure state that I was worried about the most (that the app is built correctly but then fails at runtime because xiph libs aren't installed on the user's system) is no longer possible? Xiph libs, if used, are _always_ shipped with the app?

I really think it is no longer possible! Xiph libs are always shipped when the `hooks.user_defines.flutter_soloud` is `true`. And vice versa if it is `false` or not defined.

PS: please read the changelog when v5 is published! There are almost all my milestones here :)

PR #550, opened 2026-09-03
Rename agent skills to hyphenated names for Zed compatibility

Kunstderfug:
## Problem

Zed validates agent skill names against `^[a-z0-9-]+$` and refuses to load any skill whose name contains underscores. All 14 bundled agent skills are named `flutter_soloud-*`, so Zed fails to load every one of them:

```
Skill Failed to Load
Skill name must contain only lowercase letters, numbers, and hyphens
.agents/skills/flutter_soloud-3d-audio/SKILL.md
```

(One such error per skill. Claude Code and other agents load the underscore names fine, which is why this goes unnoticed there.)

## Fix

- Rename the bundled `skills/flutter_soloud-*` directories to `skills/flutter-soloud-*` (including their `references/` subdirectories)
- Update each `SKILL.md` frontmatter `name:` to the hyphenated name
- Update sibling-skill cross-references inside skill bodies (e.g. "see the sibling skill `flutter-soloud-filters`")
- Update `test/skills_install_test.dart` to the new names

No installer changes are needed: `bin/skills.dart` derives install paths from the bundled directory names, so it picks up the new names automatically.

## Verification

- On a fresh project root, `dart run flutter_soloud:skills` installs the 14 renamed skills into `.agents/skills`, and `--check` exits 0 with "The flutter_soloud agent skills are up to date (14 installed)."
- Against a project with the old underscore-named installs, `--check` correctly reports "14 of the flutter_soloud agent skills (14 total) are not installed", so existing users will be offered the renamed set.
- `test/skills_install_test.dart` expectations updated to the hyphenated names.

## Note for existing installs

Projects that already installed the underscore-named skills keep those directories. After upgrading, `--check` reports the hyphenated skills as "not installed" (the names no longer match), and re-running the installer installs them alongside the stale ones; the old `flutter_soloud-*` directories can then simply be deleted. Happy to follow up with a small installer change that cleans up legacy underscore-named dirs during install if you'd prefer that.


alnitak:
Thank you!

PR #554, opened 2026-09-11
overhaul Xiph libraries: unified prebuilt bundling, platform-specific build hook options, and runtime diagnostics

### LINKED ISSUE #552: fix: package broken on linux/arm64 and others - don't bundle pre-built binaries in the package
`flutter_soloud` currently ships pre-built single-architecture binaries alongside the Dart source. This is the wrong pattern for a Dart package and prevents us from using this library on most of our targets.

Test environment: RaspberryPi 5 w/ Debian Trixie (linux/arm64, glibc 2.41)
Other excluded platforms:
- linux/riscv
- linux/arm64, muslc

We haven't tested Windows 11 ARM and arch variants of macOS, but this applies as well.

**Problems**

- **Single-arch only.** A binary built for `linux-x64` doesn't work on `linux-arm64` (Raspberry Pi, Apple Silicon Linux VMs), macOS, Windows, etc. Even within Linux, C ABI differences (different versions of glibc, muslc) can break things.
- **Supply chain risk.**
  - We can't audit what we're running - the binary is opaque, not diffable.
  - A compromised release pipeline can ship a malicious binary.
  - Every consumer has to trust our publisher's build environment, and it does not seem possible to audit the provenance of the binaries included in your repo

**Proposed solutions (in order of preference)**

1. (default) **Depend on a system-installed native library**: users run `brew install X` / `apt install Y`; we just `dlopen` it.
2. (fallback) **Use Dart build hooks**: (Dart 3.10+) compile the native code from source on the user's machine — self-contained, portable, no opaque artifacts.
3. (optional) **Download prebuilt libraries as a separate package**: keeps the repo source-only, the binaries should be built via CI with a proper build matrix for the possible configurations, and the build pipeline should be fully auditable (checksums, pinned dependencies, etc).

alnitak:
## Description

Closes #552
Ref https://github.com/bdero/flutter_scene/issues/386
Updated [doc](https://docs.page/alnitak/flutter_soloud_docs/get_started/xiph_libs)
Skills updated (install them using `dart run flutter_soloud:skills` or `dart run skills@ get`)

## Overview

This PR redesigns how `flutter_soloud` manages and bundles the Xiph audio libraries (`ogg`, `vorbis`, `opus`, `flac`).

The skills have been updated to reflect these changes.

Previously, prebuilt binaries and headers were scattered across individual platform directories (`android/libs`, `windows/libs`, `linux/libs`, `ios/.../libs`, `macos/.../libs`). Depending on how an app was packaged, this could lead to missing dynamic libraries or runtime crashes on end-user machines (ie linux/riscv/arm64, Win 11 arm).

With this update:
1. **Zero configuration by default**: Precompiled, tested Xiph binaries are bundled out of the box so packaged applications (Android AAB/APK, iOS IPA, macOS APP, Windows EXE, and Linux bundles) work immediately with zero external build tools required (no Git, CMake, or NDK needed).
2. **Unified directory structure**: All precompiled binaries and shared C headers are now stored in `xiph/prebuild/`.
3. **Platform-specific flexibility**: Developers can choose per-platform whether to use bundled prebuilts, link against system-installed packages (desktop), or compile directly from source via "git clone" and CMake.
4. **Actionable runtime diagnostics**: If dynamic library loading fails due to missing system packages, an actionable error message is printed with the exact installation commands for the host OS.

---

## How the Build Process Changed

- **Default (Zero Config)**: The Dart Native Assets build hook automatically links precompiled libraries from `xiph/prebuild/`.
  - **Android**: Bundles shared `.so` files into your APK/AAB under `lib/<abi>/`.
  - **iOS / macOS**: Statically links `.a` archives directly into the plugin binary, ensuring the resulting `.app` bundle is completely self-contained and never depends on external Homebrew libraries.
  - **Windows**: Bundles `.dll` files next to the compiled executable.
  - **Linux**: Bundles `.so` files alongside the application.
- **Standalone build scripts**: All helper scripts in `xiph/` (`build_Android.sh`, `build_Linux.sh`, `build_OSX.sh`, `build_iOS.sh`, `build_Windows.bat`) now output directly into `xiph/prebuild/`.
- **Compile-time status**: The build hook now logs the active target and mode during compilation (e.g. `[flutter_soloud] Building on macOS (arm64) with Xiph libs [prebuilt]`).

**Note**: Flutter redirects all Hook output to `stdout.txt` and `stderr.txt` (in the `.dart_tool/hooks_runner/flutter_soloud/<hash>/` dir) by design. The print statements in the build.dart can then be viewed only when building or running in verbose mode (`-v`).

---

## `pubspec.yaml` Configuration Options

All options are optional and configured under `hooks.user_defines.flutter_soloud` in your application's `pubspec.yaml`.

```yaml
hooks:
  user_defines:
    flutter_soloud:
      # 1. Disable all Xiph libraries across all platforms (reduces binary size)
      # Defaults to false.
      no_xiph_libs: false

      # 2. Use system-installed libraries on desktop (requires end user system to have libs installed)
      # Defaults to false.
      linux_use_system_libs: false
      macos_use_system_libs: false
      windows_use_system_libs: false

      # 3. Force building Xiph libraries from source using CMake into .dart_tool/
      # Defaults to false. Requires Git and CMake installed on your machine.
      linux_force_build_libs: false
      macos_force_build_libs: false
      windows_force_build_libs: false
      android_force_build_libs: false
      ios_force_build_libs: false
```
These defaults are harmless — no breaking changes involved.

**Note**: setting `*_force_build_libs` to true will increase the build time a lot. Compilation artifacts are cached in `flutter_soloud`'s `.dart_tool/` directory, meaning they persist even after running `flutter clean` in your application (a re-clone and re-compile only occurs if the plugin's `.dart_tool/` is removed or on fresh CI/CD environments without cache).

Tested on:
- Linux X86_64
- Windows X86_64
- MacOS silicon arm
- iOS simulator on silicon and iPhone device
- Android emulator and real device

## Type of Change

- [X] ✨ New feature (non-breaking change that adds functionality)
- [X] 📝 Documentation


alnitak:
I'll reopen this PR in another one!

alnitak:
moved to #556

